PR #95, opened 2023-09-15
fixes #90 by adding postProcess hook to scoped-css-preprocessor to break cache for associated template files

### LINKED ISSUE #90: Updating Template before CSS causes template to not pick up scoped class name
I've observed some funky behavior within the new addon, specifically around creating elements with a style class _before_ the class exists, the template never receives the scoped class definition, even after the class has been created and the template has been changed. Additional classes don't necessarily trigger a scoped class name update, only in certain cases where the element has some specific changes to it.

1. Create a template file containing a div with a single style
```hbs
// hbs
<div class="foo">foo</div>
```

2. Create an accompanying css file with the style for the div
```css
// css
.foo {
 	color: red;
 }
 ```
 
3. Observe the rendered HTML does not have a scoped class name for `foo`
```html
// html output
<div class="foo">foo</div>
```

4. Update the template file with a second div with a second single style
```hbs
// hbs
<div class="foo">foo</div>
<div class="bar">bar</div>
```

5. Observe the rendered HTML after build has applied the scoped class name for `foo`
```html
// html output
<div class="foo_ee3b6cad0">foo</div>
<div class="bar">bar</div>
```

6. Update the CSS file with a style definition for `bar`
```css
// css
.foo {
 	color: red;
 }
 .bar {
 	color: green;
 }
 ```
 
 7. Observe the rendered HTML after build to see there has been no change to the scoped class definitions.
```html
// html output
<div class="foo_ee3b6cad0">foo</div>
<div class="bar">bar</div>
```

8. Save the template file with no changes
```hbs
// hbs
<div class="foo">foo</div>
<div class="bar">bar</div>
```

 9. Observe the rendered HTML after build to see there has been no change to the scoped class definitions.
```html
// html output
<div class="foo_ee3b6cad0">foo</div>
<div class="bar">bar</div>
```

10. Add a single space character after "bar" and save the template file
```hbs
// hbs
<div class="foo">foo</div>
<div class="bar">bar </div>
```

 11. Observe the rendered HTML after build to see the first definition is scoped, but there has been no change to the second scoped class definition.
```html
// html output
<div class="foo_ee3b6cad0">foo</div>
<div class="bar">bar </div>
```

12. Add a div with no styles after the "bar" div and save the template file
```hbs
// hbs
<div class="foo">foo</div>
<div class="bar">bar </div>
<div>baz</div>
```

 13. Observe the rendered HTML after build to see there has been no change to the scoped class definitions.
```html
// html output
<div class="foo_ee3b6cad0">foo</div>
<div class="bar">bar </div>
<div>baz</div>
```
14. Add a "baz" style to the latest div and save the template file
```hbs
// hbs
<div class="foo">foo</div>
<div class="bar">bar </div>
<div class="baz">baz</div>
```

 15. Observe the rendered HTML after build to see there has been no change to the scoped class definitions.
```html
// html output
<div class="foo_ee3b6cad0">foo</div>
<div class="bar">bar </div>
<div class="baz">baz</div>
```

16. Update the CSS file with a style definition for `bar`
```css
// css
.foo {
 	color: red;
 }
 .bar {
 	color: green;
 }
.baz {
	color: blue;
}
 ```

17. Observe the rendered HTML after build to see there has been no change to the scoped class definitions.
```html
// html output
<div class="foo_ee3b6cad0">foo</div>
<div class="bar">bar </div>
<div class="baz">baz</div>

18. Add a "foo" style to the last div and save the template file
```hbs
// hbs
<div class="foo">foo</div>
<div class="bar">bar </div>
<div class="baz foo">baz</div>
```

 19. Observe the rendered HTML after build to see there has been no change to the second scoped class definition, but the third has updated.
```html
// html output
<div class="foo_ee3b6cad0">foo</div>
<div class="bar">bar </div>
<div class="baz_ee3b6cad0 foo_ee3b6cad0">baz </div>
```

20. Add a second "bar" style to the second div and save the template file
```hbs
// hbs
<div class="foo">foo</div>
<div class="bar bar">bar </div>
<div class="baz foo">baz</div>
```

 21. Observe the rendered HTML after build to see there has been no change to the scoped class definitions.
```html
// html output
<div class="foo_ee3b6cad0">foo</div>
<div class="bar bar">bar </div>
<div class="baz_ee3b6cad0 foo_ee3b6cad0">baz </div>
```

22. Add a fourth div with a "bar" style and save the template file
```hbs
// hbs
<div class="foo">foo</div>
<div class="bar bar">bar </div>
<div class="baz foo">baz</div>
<div class="bar">biz</div>
```

 23. Observe the rendered HTML after build to see there has been no change to the second scoped class definitions, but the new fourth div has generated a scoped definition.
```html
// html output
<div class="foo_ee3b6cad0">foo</div>
<div class="bar bar">bar </div>
<div class="baz_ee3b6cad0 foo_ee3b6cad0">baz </div>
<div class="bar_ee3b6cad0">biz</div>
```

The second div with `bar` never updates, even though a newer div with the same style does. Definitely appears to be some sort of caching issue, hopefully these steps help!

jakebixby:
On css changes, the postProcess hook compares the previous template build to the current one using extracted classnames that get stored between builds. If there is a difference, we append a single space to the template file, which breaks the cache for the broccoli-persistent-filter that is processing the templates (ideally we would break the cache through the filter's mechanisms, but was unable to figure it out, so this hack works for now).

### DIFF
--- a/ember-scoped-css/src/lib/scoped-css-preprocessor.js
+++ b/ember-scoped-css/src/lib/scoped-css-preprocessor.js
@@ -4,6 +4,7 @@ import Concat from 'broccoli-concat';
 import { Funnel } from 'broccoli-funnel';
 import MergeTrees from 'broccoli-merge-trees';
 import Filter from 'broccoli-persistent-filter';
+import { parseTemplates } from 'ember-template-tag';
 import path from 'path';
 
 import fsExists from './fsExists.js';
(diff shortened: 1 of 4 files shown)


ember-scoped-css/src/lib/scoped-css-preprocessor.js
```diff
@@ -7,11 +7,15 @@ import Filter from 'broccoli-persistent-filter';
...
 import fsExists from './fsExists.js';
+import { readFile, writeFile } from 'fs/promises';
 import { packageScopedPathToModulePath } from './generateAbsolutePathHash.js';
+import getClassesTagsFromCss from './getClassesTagsFromCss.js';
 import generateHash from './generateRelativePathHash.js';
 import rewriteCss from './rewriteCss.js';
+import rewriteHbs from './rewriteHbs.js';
 
 const COMPONENT_EXTENSIONS = ['hbs', 'js', 'ts', 'gjs', 'gts'];
+const TEMPLATE_EXTENSIONS = ['hbs', 'gjs', 'gts'];
```
jakebixby, line 18: will be using these for gjs/gts support later

ember-scoped-css/src/lib/scoped-css-preprocessor.js
```diff
@@ -67,12 +71,57 @@ class ScopedFilter extends Filter {
...
+          const localPackagerStylePath = packageScopedPathToModulePath(relativePath);
+          const postfix = generateHash(localPackagerStylePath);
+          const templateContents = await readFile(templateFilePath,  'utf-8');
+          const templateOriginal = rewriteHbs(templateContents, previousClasses, tags, postfix);
+          const templateRewrite = rewriteHbs(templateContents, classes, tags, postfix);
+          // if the rewrite doesn't match the original output, we want to rebuild the template
+          if (templateOriginal !== templateRewrite) {
+            // this is an awful hack because we don't know how to invalidate broccoli-persistent-filter cache
+            // ideally we'd invalidate the broccoli-peristent-filter cache here
+            await writeFile(templateFilePath, templateContents + ' ');
```
jakebixby, line 107: I feel sad that I had to revert to using a hack like this, but the cache invalidation was not working as advertised (or I was just doing it wrong). Would love to fix this to actually invalidate the cache correctly, but this will work for now.

ember-scoped-css/src/lib/scoped-css-preprocessor.js
```diff
@@ -67,12 +71,57 @@ class ScopedFilter extends Filter {
       return '';
     }
   }
+  async postProcess(results, relativePath) {
+    if (relativePath.endsWith('.module.css')) {
```
NullVoxPopuli, line 75: this should *probably* be a config option, but I think since usage is low, we can skip it

jakebixby: I'll add a config option to opt in

jakebixby: Misunderstood, will look at doing that in a later PR since `module.css` is also considered elsewhere.

However, I do think I'd like to check for CI to skip the whole thing since CI doesn't do a second pass and we'd be doing extra work for nothing.

```js
   if (process.env.CI || relativePath.endsWith('.module.css')) {
      return results;
    }
```

jakebixby:
Trying to figure out how to test this specific functionality, it's very dependent upon changing files, which isn't really in the scope of an integration or even acceptance test. Any advice would be welcome!

ember-scoped-css/src/lib/scoped-css-preprocessor.js
```diff
@@ -67,12 +72,118 @@ class ScopedFilter extends Filter {
...
+                templateContents,
+                previousClasses,
+                classes,
+                tags,
+                postfix
+              )
+            );
+          } else {
+            // find all template tags, and extract the contents to compare
+            const templates = parseTemplates(templateRaw, '');
```
jakebixby, line 124: parses templates with `ember-template-tag`, which returns an array of identified template tag objects, with `contents` representing the template markup. We can then use this for the template change check and if any did, we append a space to the file to kick the cache.

NullVoxPopuli (approved):

PR #102, opened 2023-09-26
Fix issue with sub-expressions in class strings

NullVoxPopuli:

PR #103, opened 2023-09-26
Update the v2-addon-ts package to align with the latest blueprint

NullVoxPopuli:
Ran in to an issue with floating deps over in https://github.com/soxhub/ember-scoped-css/pull/102 

and decided to take care of some other dep neglect while I was at it.

I primarily copied from the v2 addon blueprint.

Additionally, I added ember-resources to the classic and embroider test apps so that the diff in #102 is smaller. It'll be used for `cell` for a test utility.


NullVoxPopuli:
Gonna close this as it's quite a bit out of date now.

PR #111, opened 2023-10-03
Setup d.ts emitting, and clearer cross-compliing of the node assets

NullVoxPopuli:
## Notes for review

Lockfile: 
image

also,
- node-tests moved inline, unchanged though
- find-up had to get downgraded because we can't actually use the real package if we don't convert it to cjs (find-up@v6+ is ESM-only)

## What happened?

While working on a bug, I noticed that I needed types for `scopedClass`.

One thing lead to another, and I ended up with this.
- node/esm compiled to its own folder
- node/cjs compiled to its own folder
- runtime stuff compiled to its own  folder
- using externals (faster build, faster install)
  - we no longer convert all dependencies to cjs and bundle them with our files
- switch away from mocha to vitest's in-source testing so that we can still test private things, but not needlessly expose those private things through package.json#exports (this also needed to happen because we have a build going now, and vitest "just works":tm:)
- JSDoc for `scopedClass` (and an overload)

image

By using externals all of the modules are smaller, less complex, and easier to parse / cache (and share cache with other tooling)


ember-scoped-css/build.mjs
NullVoxPopuli, line 1: build.sh was simultaneously getting too complex and had no ergonomic way to share data between the different build configs.

The entire build
- 3 esbuild invocations
- 1 tsc invocation

takes 0.9s

ember-scoped-css/package.json
```diff
@@ -14,23 +14,29 @@
     "addon-main.cjs"
   ],
   "exports": {
```
NullVoxPopuli, line 16: changes here now reflect the `dist` directory, and everything references the built output, rather than some things being `src`, and some not

ember-scoped-css/src/build/app-css-livereload-loader.js
```diff
@@ -1,5 +1,6 @@
-import { readFile } from 'fs/promises';
-import path from 'path';
+import { readFile } from 'node:fs/promises';
```
NullVoxPopuli, line 1: I converted some node imports to have the `node:` prefix

ember-scoped-css/src/lib/generateAbsolutePathHash.js
```diff
@@ -1,9 +1,9 @@
 import fsSync from 'node:fs';
 import path from 'node:path';
 
-import { findUpSync } from 'find-up';
+import findUp from 'find-up';
```
NullVoxPopuli, line 4: find-up's api changed from v6 -> v5

ember-scoped-css/src/lib/generateRelativePathHash.ts
```diff
@@ -1,5 +1,5 @@
 import md5 from 'blueimp-md5';
 
-export default function generateRelativePathHash(relativePath) {
+export default function generateRelativePathHash(relativePath: string) {
```
NullVoxPopuli, line 3: this was converted to TS because I wanted `scopedClass` to be TS, so I could provide an overload type for it.

ember-scoped-css/src/lib/renameClass.ts
NullVoxPopuli, line 1: this was converted to TS because I wanted scopedClass to be TS, so I could provide an overload type for it.

ember-scoped-css/src/runtime/test-support.ts
NullVoxPopuli, line 1: the bulk of the meaningful changes are here:
- overload function signature
- lots of JSDoc
- scopedClass now supports not passing a class, so you can _just_ grab the hash -- this is useful for testing plain elements, which we couldn't do today

jacojoubert (approved):

mattmcmanus (approved):
Love it. Awesome work.

PR #116, opened 2023-10-04
Pods support

NullVoxPopuli:
Pods was not previously supported (you have to come up with a name for the style file, and that was the missing piece).

Adding pods support for embroider using apps wasn't able to complete because we can't safely do transforms on the wireformat. We need to be using `transforms` from: https://github.com/emberjs/babel-plugin-ember-template-compilation/#common-options that would allow us to use the existing template-transform code from the classic app side of things and get rid of all the current wire-format mangling. 
The wire format is not public API to us, and we can't ever expect that it's stable.

This is now the second feature / fix I've not been able to bring pairity to embroider due to the wire format stuff. The other was: https://github.com/soxhub/ember-scoped-css/pull/102/files#diff-6d080da1153c9125b2c788dbaf89ade7aff1156210836bf68adbd3c6c032f21fR22


ember-scoped-css/src/build/app-js-unplugin.js
```diff
@@ -63,8 +63,22 @@ export default createUnplugin(({ appDir }) => {
      * @returns
      */
     async transform(code, id) {
-      const cssPath = id.replace(/(\.js)|(\.hbs)/, '.css');
-      const postfix = generateHash(cssPath);
+      let cssPath = id.replace(/(\.js)|(\.hbs)/, '.css');
+      let moduleGroupPath = cssPath;
+
+      /**
+       * Pods support
```
NullVoxPopuli, line 70: pods support is ultimately these type of code littered everywhere

ember-scoped-css/src/lib/replaceGlimmerAst.js
```diff
@@ -10,7 +10,7 @@ const parseOptions = {
   parser: babelParser,
 };
 
-export default async function (script, id, replaceFunction) {
+export default async function replaceGlimmerAst(script, id, replaceFunction) {
```
NullVoxPopuli, line 13: gave this a name so I could find it better

ember-scoped-css/src/lib/scoped-css-preprocessor.js
```diff
@@ -82,15 +92,26 @@ class ScopedFilter extends Filter {
     for (let inputPath of this.inputPaths) {
       const templateFile = {};
 
-      for (let ext of TEMPLATE_EXTENSIONS) {
+      eachExtension: for (let ext of TEMPLATE_EXTENSIONS) {
```
NullVoxPopuli, line 95: label for being clear about which for loop I'm breaking out of

ember-scoped-css/src/lib/scoped-css-preprocessor.js
```diff
@@ -82,15 +92,26 @@ class ScopedFilter extends Filter {
...
+          let podsName = relativePath.replace(/styles\.css$/, 'template.hbs');
+
+          templateFilePath = path.join(inputPath, podsName);
+          exists = await fsExists(templateFilePath);
+        }
 
         if (exists) {
           templateFile.path = templateFilePath;
           templateFile.ext = ext;
-          continue;
```
NullVoxPopuli, line 93: continue here, I believe was a mistake. once we find a file, we don't want to keep searching for more files.

ember-scoped-css/src/lib/scoped-css-preprocessor.js
```diff
@@ -196,17 +217,28 @@ export default class ScopedCssPreprocessor {
...
-      include: [
-        this.appName + '/components/**/*.css',
-        ...COMPONENT_EXTENSIONS.map(
-          (ext) => this.appName + `/components/**/*.${ext}`,
-        ),
-      ],
-    });
+    let roots = [
+      this.appName + '/components/',
+      ...(this.userOptions.additionalRoots || []).map(
```
NullVoxPopuli, line 226: this ended up being the primary motivator for the `additionalRoots` option -- we had a few places that hardcoded to searching for components.

atm, `additionalRoots` is only needed for classic builds.

ember-scoped-css/src/scoped-babel-plugin.js
```diff
@@ -31,7 +38,20 @@ export default () => {
             relativeFileName,
           );
 
-          const cssPath = fileName.replace(/(\.js)|(\.ts)/, '.css');
+          let cssPath = fileName.replace(/(\.js)|(\.ts)/, '.css');
+
+          /**
```
NullVoxPopuli, line 43: as you can see, the impact of pods is _immense_

test-apps/pods-embroider-app/package.json
```diff
@@ -45,17 +45,14 @@
...
-    "ember-data": "~5.3.0",
-    "ember-fetch": "^8.1.2",
     "ember-load-initializers": "^2.1.2",
     "ember-modifier": "^4.1.0",
     "ember-page-title": "^8.0.0",
     "ember-qunit": "^8.0.1",
     "ember-resolver": "^11.0.1",
     "ember-scoped-css": "workspace:^",
-    "ember-source": "~5.3.0",
+    "ember-source": "~4.10.0",
```
NullVoxPopuli, line 54: I had changed this to the same version as the other embroider app in hopes the wire format would be the same as what was expected during initial implementation.

I think I may be getting trolled by a floating dependency, but I'm not sure.

It's all going to go away soon anyway.

test-apps/pods-embroider-app/package.json
```diff
@@ -45,17 +45,14 @@
     "ember-cli-dependency-checker": "^3.3.2",
     "ember-cli-htmlbars": "^6.3.0",
     "ember-cli-inject-live-reload": "^2.1.0",
-    "ember-data": "~5.3.0",
```
NullVoxPopuli, line 48: these dependencies aren't needed, and made debugging more strenuous when having breakpoints all over the place

test-apps/pods-embroider-app/package.json
```diff
@@ -20,7 +20,7 @@
     "lint:js": "eslint . --cache",
     "lint:js:fix": "eslint . --fix",
     "start": "ember serve",
-    "test": "pnpm test:ember",
+    "test": "echo 'Would run \"pnpm test:ember\", but we need a different approach for transforming templates under embroider (like a template-transpilation plugin, see: https://github.com/emberjs/babel-plugin-ember-template-compilation/#common-options)'",
```
NullVoxPopuli, line 23: tests disabled for now

NullVoxPopuli:
> EADDRINUSE

looks like I need to get all the test apps their own port (port 0)

jakebixby (approved):

PR #140, opened 2024-03-04
fix TS compatibility / add Glint support

NullVoxPopuli:
## Changes

- Adds a `template-registry.d.ts` so that we can have ember-scoped-css work with Glint without custom definitions

## For review

You'll probably want to collapse the test directories first:

image


ember-scoped-css/template-registry.d.ts
NullVoxPopuli, line 1: Here along with the package.json change is the actual fix

ember-scoped-css/package.json
```diff
@@ -11,9 +11,14 @@
   "files": [
     "src",
     "dist",
-    "addon-main.cjs"
+    "addon-main.cjs",
+    "template-registry.d.ts"
   ],
   "exports": {
+    "./template-registry": {
+      "types": "./template-registry.d.ts",
```
NullVoxPopuli, line 19: the actual fix

NullVoxPopuli:
Seems maybe there is a real problem. Hmm

mattmcmanus (approved):

NullVoxPopuli:
Looks like this accidentally got blocked by: https://github.com/soxhub/ember-scoped-css/issues/108

PR #147, opened 2024-03-07
Support gjs and gts / strict usage -- `scopedClass` needs an import (that is then removed at build)

NullVoxPopuli:
Resolves: https://github.com/soxhub/ember-scoped-css/issues/146


.changeset/quiet-icons-think.md
```diff
@@ -0,0 +1,15 @@
...
+
+Support `gjs` and `gts` (`<template>`) by allowing an import for `scopedClass`.
+
+Example usage:
+
+```gjs
+import { scopedClass } from 'ember-scoped-css';
+
+<template>
+  <p class={{scopedClass "hi"}}>text color</p>
```
wagenet, line 13: Do you _need_ to do this? Or is it just allowed now?

wagenet (approved):

.changeset/quiet-icons-think.md
NullVoxPopuli: we current use `scoped-class` in the template, and that works, but we have to disable linting and type checking in the template to do it.

This changes allows us to keep linting / type checking enabled.

PR #159, opened 2024-03-27
Support importing `scopedClass` in gjs/gts in classic-apps

NullVoxPopuli:
Continues the work of https://github.com/soxhub/ember-scoped-css/pull/147,
- adding the ability to import `scopedClass` in classic apps.
- fixes an issue where the runtime module for `ember-scoped-css` was not included in the `dist` directory


ember-scoped-css/build.mjs
```diff
@@ -55,7 +55,7 @@ await esbuild.build({
 
 // Runtime
 await esbuild.build({
-  entryPoints: ['src/runtime/test-support.ts'],
+  entryPoints: ['src/runtime/test-support.ts', 'src/runtime/index.ts'],
```
NullVoxPopuli, line 58: part 1 of the fix is here, the runtime module is required to be resolved, even if never used

jacojoubert (approved):

ember-scoped-css/src/scoped-babel-plugin.js
```diff
@@ -6,22 +6,78 @@ import generateHash from './lib/generateRelativePathHash.js';
...
+   *   - the babel-plugin-ember-template-compilation step has not run yet,
+   *     else we'd see precompileTemplate and setComponentTemplate (and more imports)
+   *
+   *   - note that in ember-template-imports' implementation, the file changes
+   *     after `ImportDeclaration` visitors have ran, and by the time we see
+   *     CallExpressions, we have the familiar `setComponentTemplate`
+   */
   return {
     visitor: {
+      ImportDeclaration(path, state) {
```
NullVoxPopuli, line 44: part 2 of the fix is here, we need to remove the import -- this is mostly duplicate code, but any overlap between the AST structures of Babel and Recast seems to be cooincidence, and I couldn't find anything saying they must be the same. 
The two AST visitor tools also have different APIs for removing / entering nodes -- an abstraction could be made -- not sure if worth it yet.

PR #174, opened 2024-05-10
Proper Embroider Support

NullVoxPopuli:
This will cause a change in documentation as it'll be far more stable (and also our way of fixing some embroider compatibility problems due to messing with the wire format).

Breaking changes:
- we no longer use the appJsUnplugin in embroider
- no longer supported:
  - ember-template-imports earlier than v4
  - babel-plugin-ember-template-compilation earlier than 2.2.5
  - rollup-plugin-glimmer-template-tag


NullVoxPopuli:
This has been resolved in scoped-css @  0.21.0

PR #225, opened 2024-06-11
Fix mishandling of package lookup cache

NullVoxPopuli:
the cache ended up being too hard to get correct, and the perf saved wasn't worth it

### DIFF
--- a/ember-scoped-css/src/lib/path/utils.js
+++ b/ember-scoped-css/src/lib/path/utils.js
@@ -1,3 +1,4 @@
+import assert from 'node:assert';
 import fsSync from 'node:fs';
 import path from 'node:path';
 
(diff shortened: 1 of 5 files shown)


ember-scoped-css/src/lib/path/utils.js
```diff
@@ -122,7 +123,7 @@ const UNSUPPORTED_DIRECTORIES = new Set(['tests']);
  * @param {string[]} [additionalRoots]
  * @returns
  */
-export function isRelevantFile(fileName, additionalRoots) {
+export function isRelevantFile(fileName, { additionalRoots, cwd }) {
```
NullVoxPopuli, line 126: cwd is now passed in, as it's too hard to mock globals or even modules

PR #285, opened 2025-03-19
add support for interop with other pipelines

mansona:
This should allow us to interop better with other css pipelines (such as ember-css-modules wile people are migrating). In my testing I am able to pass the following config to the ember app: 

```js
    cssModules: {
      extension: 'module.css',
      intermediateOutputPath: 'super-css-modules.css',
    },
    'ember-scoped-css': {
      passthrough: ['super-css-modules.css'],
      passthroughDestination: 'assets',
    },
```

and then I am able to use 

```css
@import './super-css-modules.css';
```

in my app.css and I can have both system working together 🎉

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,6 +86,8 @@ const app = new EmberApp(defaults, {
   'ember-scoped-css': {
     layerName: 'app-styles', // default: 'components', set to false to disable the layer
     additionalRoots: ['routes/'], // default: [], set this to use scoped-css in pods-using apps
+    passthrough: ['some-other-file.css'], // default: [] this is only used in a non-embroider app to pass files through the build pipeline
+    passthroughDestination: 'assets' // default: undefined this alters where the passthrough files are placed in the output tree
   }
 });
 ```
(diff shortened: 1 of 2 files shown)


ember-scoped-css-compat/index.js
```diff
@@ -59,8 +59,9 @@ module.exports = {
     let scopedCssOptions = this.app.options['ember-scoped-css'] || {};
 
     // we need to run our css preprocessor first, so we removed all other from the registry and run them as part of our preprocessor
-    // we did not find other way how to do it more elegantly
-    let preprocessors = registry.load('css');
+    // we did not find other way how to do it more elegantly. Also in later ember-cli versions this array becomes a live reference
+    // so removing them from the registry actually removes them from this reference 😱 so we need to make a deep copy here
+    let preprocessors = [...registry.load('css')];
```
mansona, line 64: I was originally testing in ember 3.28 but something changed since then that required that I do this 🫠

PR #291, opened 2025-03-26
Drop support for Node <20 and use package-up instead of custom function

mansona:
I was getting the following error during github CI: 

```
Could not determine project for /path/to/app/app/templates/index.gjs
```

And I figured it was some specific problem with Linux and the custom `findPackageJsonUp` function. Instead of trying to fix every iteration of this function on every OS I thought it was better to just use a tried and tested `package-up` 👍 

One caveat: this now requires Node 20+ because package-up is ESM and you need to rely on `require() ESM` for this


NullVoxPopuli (approved):

NullVoxPopuli:
It's the best when code can be deleted and the tests don't notice. haha

PR #303, opened 2025-05-30
Add compat-embroider + vite test-app and some tests for things with hbs under vite

NullVoxPopuli:


NullVoxPopuli:
testing this out locally atm

PR #313, opened 2025-07-16
Internalize md5

NullVoxPopuli:
bundling / reducing runtime dependencies that vite has to process for runtime is better / faster / less complex for consumers.


ember-scoped-css/package.json
```diff
@@ -79,22 +79,17 @@
   "dependencies": {
     "@babel/parser": "^7.26.9",
     "@embroider/addon-shim": "^1.9.0",
-    "acorn": "^8.14.1",
```
NullVoxPopuli, line 82: I also removed all dependencies that weren't imported

Issue #270, opened 2024-10-31
css keywords are being incorrectly scoped


runspired:
rules like `:nth-of-type(even)` use keywords like `even` and `odd` or numbers. Currently the class hash is being attached to this:

Example:

![image](https://github.com/user-attachments/assets/ff645c47-d5cb-4077-b0af-f23a79e632dd)

PR #338, opened 2025-10-15
Don't add scoping postfix to contents of `nth-of-type` psuedo-selector

NullVoxPopuli:
Fixes: #270


ember-scoped-css/src/lib/css/rewrite.js
```diff
@@ -59,56 +79,3 @@ export function rewriteCss(css, postfix, fileName, layerName) {
     '\n}\n'
   );
 }
-
-if (import.meta.vitest) {
```
NullVoxPopuli: extracted to a real file

Issue #356, opened 2025-10-31
Separate CSS file not reloading on change


gossi:
I've run into a case, where I had a component in `templates/` folder (`src/templates/user/-components/user-agent.{gts,css}`). I noticed, there are two issues with that.

## 1. Double the File

I saw a strange behavior in which the CSS file showed up twice in `<head>`:

<img width="815" height="68" alt="Image" src="https://github.com/user-attachments/assets/527c6c8e-c1dc-4562-8c4f-74b84a142b8a" />

When I made a change, then only the unprocessed file `user-agent.css` was changed, but that was also the wrong file. Here I found the culprit real quick:

```ts
// src/app.ts

export default class App extends EmberApp {
  modules = {
    // ...
    ...import.meta.glob('./templates/**/*', { eager: true })
  };
}
```

that can be fixed with:

```
// src/app.ts

export default class App extends EmberApp {
  modules = {
    // ...
    ...import.meta.glob(['./templates/**/*', '!./templates/**/*.css'], { eager: true })
  };
}
```

that excludes CSS files, that are considered for scoped usage (all for me). That needs documentation.

## 2. Proceesed file does not change Contents

But the file processed by `ember-scoped-css` is not refreshing. I see the import statement is injected into the transpiled component gts file:

<img width="992" height="475" alt="Image" src="https://github.com/user-attachments/assets/13b86f19-4ada-4307-a2ca-1788764f48c1" />

And the hashed css file, that got all the vite snippet for hmr in it:

<img width="1311" height="473" alt="Image" src="https://github.com/user-attachments/assets/7c8dc96b-6e27-49d1-9a22-9a4bb482557e" />

... but that does not HMR 😬 The browser is reloading, but the contents of the file stay the same. Can you nudge me, which code function from `ember-scoped-css` is responisble for that, then I can take a look at fixing that.

PR #357, opened 2025-10-31
Address issue about import.meta.glob mentioned in #356

NullVoxPopuli:


README.md
```diff
@@ -87,6 +87,43 @@ export default defineConfig({
...
+  ```
+
+  ✅ Do this
+  ```js
+  ...import.meta.glob('./templates/**/*.{gjs,gts}', { eager: true }),
+  ```
+  or better yet, for small projects:
+  ```js
+  ...import.meta.glob('./templates/{top,level,folders}/{sub,folders}.{gjs,gts}', { eager: true }),
+  ```
```
gossi: Looks like the correct fix, but in the wrong location: https://github.com/ember-cli/ember-strict-application-resolver#usage - this is where the copypasta is coming from.

PR #359, opened 2025-11-02
Fix reload problem with separate CSS files

### LINKED ISSUE #356: Separate CSS file not reloading on change

gossi:
Fixes #356

Also it improves the readability for `<link>` in head during development for better debugging. It changes the `href`:

- For `<style>`: `/path/to/component-[hash].css`
- For separate CSS File: `/path/to/component.css?scoped=[hash]`

The reason being vite handles CSS files itself with HMR already. When the URL is changed (different resolve), then the original file is no longer watched. The idea is to keep the original URL with a `?scoped=[hash]` then use the `transform()` hook in unplugin to rewrite the CSS.

That also means, the unplugin needs the config, that was previously only needed for the template plugin.

### DIFF
--- a/ember-scoped-css/src/build/unplugin-colocated.js
+++ b/ember-scoped-css/src/build/unplugin-colocated.js
@@ -51,6 +51,9 @@ export function colocated(options = {}) {
           path.basename(parsed.fileName),
         );
 
+        /**
+         * Rollup doesn't normally watch CSS files
+         */
         this.addWatchFile(filePath);
 
         return buildResponse(id, filePath);
@@ -74,14 +77,25 @@ export function colocated(options = {}) {
     },
     vite: {
       /**
-       * @param {string} id location on disk
+       * There may not be meta for this request yet.
+       *
+       * @param {*} id
        */
       load(id) {
         if (request.is.colocated(id)) {
-          let filePath = id.split('?')[0];
-          let response = buildResponse(id, filePath);
+          const parsed = request.colocated.decode(id);
 
-          return response.meta[META].css;
+          let code = readFileSync(parsed.fileName, 'utf-8');
+          let relativeFilePath = path.relative(CWD, parsed.fileName);
+
+          let css = rewriteCss(
+            code,
+            parsed.postfix,
+            relativeFilePath,
+            options.layerName,
+          );
+
+          return css;
         }
       },
     },


ember-scoped-css/src/build/scoped-css-unplugin.js
```diff
@@ -1,39 +1,130 @@
...
+      // vite: transform separate CSS file
+      if (isSeparateCSSFileRequest(id)) {
+        const parsed = decodeSeparateCSSFileRequest(id);
+        const filePath = meta?.fileName ?? path.relative(cwd(), id);
+
+        return rewriteCss(code, parsed.postfix, filePath, options.layerName);
+      }
+    },
+
+    // this below is actually the format for transforms, so rust based runtimes
```
NullVoxPopuli: leftover comment

ember-scoped-css/src/build/scoped-css-unplugin.js
```diff
@@ -1,39 +1,130 @@
...
 /**
- * The plugin that handles CSS requests from other transform (e.g.: babel)
- *
- * This plugin takes no options because the CSS transform has already happened by the time
- * we handle it here.
+ * The plugin that handles CSS requests for `<style>` elements and transforms
+ * for existing files
  */
-export default createUnplugin(() => {
+export default createUnplugin((options = {}) => {
```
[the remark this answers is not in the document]
NullVoxPopuli: Added to this PR

ember-scoped-css/src/build/scoped-css-unplugin.js
```diff
@@ -1,39 +1,122 @@
...
+          );
+          const relativeFilePath = path.relative(CWD, filePath);
+          
+          let code = readFileSync(filePath, 'utf-8');
+          let css = rewriteCss(
+            code,
+            parsed.postfix,
+            relativeFilePath,
+            options.layerName,
+          );
```
gossi: you wanna leave that in `transform()` - as this is what's happening. More of a semantic thing, but nevertheless

PR #360, opened 2025-11-03
Unify Public Build API

gossi:
... my inner monk couldn't handle the different ways to import the same things. So I unified them.

That is under the assumption, there is a breaking change anyway and no deprecations are needed.


ember-scoped-css/README.md
```diff
@@ -68,12 +70,24 @@ This build tool can emit CSS in a `@layer`.
 ## Installation
 
 ```bash
-npm install --save-dev ember-scoped-css
+pnpn add -D ember-scoped-css
```
NullVoxPopuli: Why make unneeded changes? <3

PR #366, opened 2025-11-07
Experimental support for `rolldown`/`tsdown`

gossi:
rolldown doesn't support CSS handling yet. There is a special handling needed, until it is supported by rolldown itself.

This PR brings this special handling for rolldown only.

### DIFF
--- a/test-apps/v2-addon-tsdown/src/index.js
+++ b/test-apps/v2-addon-tsdown/src/index.js
@@ -1,7 +1,7 @@
+export { default as Body } from './card.gjs';
 export { default as Alert } from './components/alert.gjs';
 export { default as Footer } from './components/footer.gjs';
 export { default as Header } from './components/header.gjs';
-export { default as Body } from './card.gjs';
 export {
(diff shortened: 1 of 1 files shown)


ember-scoped-css/package.json
```diff
@@ -64,6 +68,7 @@
     "test": "vitest"
   },
   "dependencies": {
+    "@ast-grep/napi": "^0.39.9",
```
gossi: I had this as devDep at first, but this is installing the binaries for each os. That's why then the build was failing (for `ember-scoped-css`).

The alternative is to make it a peer dep. Then optional? But `rolldown` users have to install it.

NullVoxPopuli:
is any of this still needed since rolldown supports rollup plugins? 🤔

gossi:
I switched my local packages to v2.2.0 and it works with rolldown, too.

We can close this here, but I wouldn't consider the case closed as of yet. There is quite some things happening in rolldown/tsdown land when it comes to CSS.

PR #382, opened 2026-03-20
Add CSS preprocessor support (SCSS/Sass/Less/Stylus)

evoactivity:
I let claude have a bash at this to get the ball rolling.

---

## Add CSS preprocessor support (SCSS/Sass/Less/Stylus)
Enables `<style scoped lang="scss">` (and other preprocessor languages) in Glimmer component templates. Preprocessing is handled by Vite's built-in `preprocessCSS` API, so **no new mandatory dependencies** are introduced — you just need the preprocessor already installed in your app (e.g. `sass`).

**Vite only.** Rollup and webpack builds will throw a clear error if `lang=` is used.

### How it works
**Babel stage (class/tag extraction)**
`getCSSContentInfo` now accepts an optional `lang` parameter. When `lang` is `scss` or `sass`, it parses with `postcss-scss` as a tolerant syntax parser — this lets the Glimmer AST plugin extract scoped class/tag names from SCSS without crashing on nesting, `&` selectors, or variables. The hash is still computed from the raw pre-compiled source.

**Vite stage (actual compilation)**
Both `unplugin-inline` and `unplugin-colocated` capture Vite's `preprocessCSS` function during `configResolved` (resolved via `createRequire` from the app root, so the correct Vite installation is always used). Their `load` hooks are now async and call `preprocessCSS` before passing the result to `rewriteCss`.
For inline styles, the `lang` value is encoded as a `&lang=<value>` query parameter on the virtual module URL, so it survives the round-trip from Babel to Vite's plugin system.

**`<style scoped inline lang="...">` downgrade**
Preprocessing is async and cannot run at Babel-time. If `inline` and `lang` are both present, the plugin:
1. Logs a warning explaining the limitation
2. Removes the `<style>` tag from the template (same as non-inline)
3. Emits a virtual CSS module import that gets preprocessed at Vite load time

### Changes

- **`ember-scoped-css/package.json`** — adds `postcss-scss` as a dependency (used as a tolerant parser at Babel-time, not a compiler)
- **`src/lib/css/utils.js`** — `getCSSContentInfo(css, lang?)` uses `postcss-scss` syntax when `lang` is `scss`/`sass`
- **`src/lib/request.js`** — `inline.create` accepts an optional `lang` param encoded into the virtual URL; `inline.decode` returns it
- **`src/build/template-plugin.js`** — reads `lang` attribute, passes it through to `getCSSContentInfo` and `request.inline.create`; handles `inline + lang` downgrade
- **`src/build/unplugin-inline.js`** — `configResolved` captures `preprocessCSS`; `load` is now async and calls it when `lang` is set
- **`src/build/unplugin-colocated.js`** — same pattern; `vite.load` preprocesses colocated files whose extension is in `{.scss, .sass, .less, .styl, .stylus}`

### Tests

**Unit (`template-plugin.test.ts`)**
- `<style scoped lang="scss">` — style tag removed, virtual import emitted with `&lang=scss`
- `<style scoped inline lang="scss">` — same output + warning logged (downgrade behaviour)

**Unit (`utils.js`)**
- `getCSSContentInfo` with SCSS nesting parses without throwing for `lang=scss` and `lang=sass`

**Integration (`test-apps/vite-app`)**
- Variables, nesting, `&` selectors
- Inline `@mixin` / `@include` with arguments
- `@use './mixins' as m` — imports a colocated `_mixins.scss` partial and calls its mixins; validates that class scoping and compiled styles are both correct

### DIFF
--- a/ember-scoped-css/src/build/template-plugin.js
+++ b/ember-scoped-css/src/build/template-plugin.js
@@ -127,38 +127,24 @@ export function createPlugin(config) {
 
             addInfo(info);
 
-            /**
-             * For <style scoped inline lang="..."> we cannot preprocess at Babel-time
-             * (preprocessing is async and requires Vite's ResolvedConfig).
-             * Downgrade to non-inline: remove the tag and inject via virtual CSS module.
-             */
-            if (hasInlineAttribute(styleTag) && lang) {
+            if (hasInlineAttributeWithoutLang(styleTag)) {
+              /**
+               * This will be handled in ElementNode traversal
+               */
+              return;
+            }
+
+            if (lang) {
+              /**
+               * For <style scoped inline lang="..."> we cannot preprocess at Babel-time
+               * (preprocessing is async and requires Vite's ResolvedConfig).
+               * Remove the tag and inject via virtual CSS module and warn user.
+               */
               console.warn(
                 `[ember-scoped-css] <style scoped inline lang="${lang}"> is not supported ` +
                   `(preprocessing is async and cannot run at Babel-time). ` +
                   `Downgrading to non-inline: the style tag will be removed and injected as a virtual CSS module.`,
               );
-
-              let cssRequest = request.inline.create(
-                info.id,
-                postfix,
-                css,
-                lang,
-              );
-
-              env.meta.jsutils.importForSideEffect(cssRequest);
-
-              // Signal ElementNode visitor to remove the tag
-              styleTag.__downgradeInline = true;
-
-              return;
-            }
-
-            /**
-             * This will be handled in ElementNode traversal
-             */
-            if (hasInlineAttribute(styleTag)) {
-              return;
             }
 
             let cssRequest = request.inline.create(info.id, postfix, css, lang);
@@ -183,12 +169,7 @@ export function createPlugin(config) {
               );
             }
 
-            // Downgraded inline+lang: remove the tag (import was already emitted in Template visitor)
-            if (node.__downgradeInline) {
-              return null;
-            }
-
-            if (hasInlineAttribute(node)) {
+            if (hasInlineAttributeWithoutLang(node)) {
               let text = textContent(node);
               let scopedText = rewriteCss(
                 text,
@@ -209,7 +190,7 @@ export function createPlugin(config) {
             return null;
           }
 
-          if (hasInlineAttribute(node)) {
+          if (hasInlineAttributeWithoutLang(node)) {
             throw new Error(
               `<style inline> is not valid. Please add the scoped attribute: <style scoped inline>`,
             );
@@ -243,10 +224,14 @@ function hasScopedAttribute(node) {
   );
 }
 
-function hasInlineAttribute(node) {
+function hasInlineAttributeWithoutLang(node) {
   if (!node) return;
   if (node.tag !== 'style') return;
   if (node.type !== 'ElementNode') return;
+
+  if (getLangAttribute(node)) {
+    return false;
+  }
 
   return node.attributes.some(
     (attribute) => attribute.name === INLINE_ATTRIBUTE_NAME,
--- a/ember-scoped-css/src/build/template-plugin.js
+++ b/ember-scoped-css/src/build/template-plugin.js
@@ -122,9 +122,37 @@ export function createPlugin(config) {
 
           if (hasScopedAttribute(styleTag)) {
             let css = textContent(styleTag);
-            let info = getCSSContentInfo(css);
+            let lang = getLangAttribute(styleTag);
+            let info = getCSSContentInfo(css, lang);
 
             addInfo(info);
+
+            /**
+             * For <style scoped inline lang="..."> we cannot preprocess at Babel-time
+             * (preprocessing is async and requires Vite's ResolvedConfig).
+             * Downgrade to non-inline: remove the tag and inject via virtual CSS module.
+             */
+            if (hasInlineAttribute(styleTag) && lang) {
+              console.warn(
+                `[ember-scoped-css] <style scoped inline lang="${lang}"> is not supported ` +
+                  `(preprocessing is async and cannot run at Babel-time). ` +
+                  `Downgrading to non-inline: the style tag will be removed and injected as a virtual CSS module.`,
+              );
+
+              let cssRequest = request.inline.create(
+                info.id,
+                postfix,
+                css,
+                lang,
+              );
+
+              env.meta.jsutils.importForSideEffect(cssRequest);
+
+              // Signal ElementNode visitor to remove the tag
+              styleTag.__downgradeInline = true;
+
+              return;
+            }
 
             /**
              * This will be handled in ElementNode traversal
@@ -133,7 +161,7 @@ export function createPlugin(config) {
               return;
             }
 
-            let cssRequest = request.inline.create(info.id, postfix, css);
+            let cssRequest = request.inline.create(info.id, postfix, css, lang);
 
             env.meta.jsutils.importForSideEffect(cssRequest);
           }
@@ -153,6 +181,11 @@ export function createPlugin(config) {
               throw new Error(
                 '<style scoped> tags must be at the root of the template, they cannot be nested',
               );
+            }
+
+            // Downgraded inline+lang: remove the tag (import was already emitted in Template visitor)
+            if (node.__downgradeInline) {
+              return null;
             }
 
             if (hasInlineAttribute(node)) {
@@ -198,6 +231,7 @@ export function createPlugin(config) {
  */
 const SCOPED_ATTRIBUTE_NAME = 'scoped';
 const INLINE_ATTRIBUTE_NAME = 'inline';
+const LANG_ATTRIBUTE_NAME = 'lang';
 
 function hasScopedAttribute(node) {
   if (!node) return;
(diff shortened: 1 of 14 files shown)


ember-scoped-css/src/build/template-plugin.js
```diff
@@ -122,18 +122,46 @@ export function createPlugin(config) {
...
+            let info = getCSSContentInfo(css, lang);
 
             addInfo(info);
 
+            /**
+             * For <style scoped inline lang="..."> we cannot preprocess at Babel-time
+             * (preprocessing is async and requires Vite's ResolvedConfig).
+             * Downgrade to non-inline: remove the tag and inject via virtual CSS module.
+             */
+            if (hasInlineAttribute(styleTag) && lang) {
```
NullVoxPopuli: alternative, hasInlineAttribute could return false when lang is present -- then I think the existing request / importForSideEffect could be used?

NullVoxPopuli: and then this wouldn't be needed https://github.com/auditboard/ember-scoped-css/pull/382/changes#diff-89557702bf5eae0554341eb941fbe5fa144154674a1987af871f4959b0ce004bR186

ember-scoped-css/src/build/template-plugin.test.ts
```diff
@@ -194,3 +194,66 @@ it('scoped inline transforms correctly', async () => {
...
+
+    // The style tag should be removed (it's a virtual module, not inline)
+    expect(templateContentsOf(output)).toMatchInlineSnapshot(`
+      [
+        "<div class="foo_e65d154a1">hi</div>",
+      ]
+    `);
+
+    // The virtual module import should include &lang=scss
+    expect(output).toContain('lang=scss');
```
NullVoxPopuli: this output is pretty big -- is there a way we can narrow it down to mean the URL contains lang=scss?

ember-scoped-css/src/build/unplugin-colocated.js
```diff
@@ -73,18 +89,57 @@ export function colocated(options = {}) {
...
     },
     vite: {
+      async configResolved(config) {
+        viteConfig = config;
+
+        // Resolve Vite's preprocessCSS from the app root to ensure we find
+        // the correct Vite installation (not a stale or missing one).
+        try {
+          const require = createRequire(
+            path.resolve(config.root, 'package.json'),
```
NullVoxPopuli: why not CWD?

ember-scoped-css/src/build/unplugin-colocated.js
```diff
@@ -73,18 +89,57 @@ export function colocated(options = {}) {
...
           const parsed = request.colocated.decode(id);
 
           let code = readFileSync(parsed.fileName, 'utf-8');
           let relativeFilePath = path.relative(CWD, parsed.fileName);
 
+          const ext = path.extname(parsed.fileName).toLowerCase();
+
+          if (PREPROCESSED_EXTENSIONS.has(ext)) {
+            if (!viteConfig || !preprocessCSS) {
+              throw new Error(
```
NullVoxPopuli, line 126: thanks for the error! 🎉

ember-scoped-css/src/lib/css/utils.js
```diff
@@ -85,4 +90,28 @@ if (import.meta.vitest) {
     expect(tags.size).to.equal(1);
     expect([...tags]).to.have.members(['div']);
   });
+
+  it('should parse SCSS nesting syntax without crashing when lang=scss', function () {
+    const scss = `
```
NullVoxPopuli, line 129: can we put non-standards CSS, scss in here? this would parse as normal CSS as is

ember-scoped-css/src/lib/css/utils.js
```diff
@@ -85,4 +90,28 @@ if (import.meta.vitest) {
...
+
+  it('should parse SCSS nesting syntax without crashing when lang=sass', function () {
+    const scss = `
+      .block {
+        &--modifier { color: green; }
+      }
+    `;
+    const { classes } = getCSSContentInfo(scss, 'sass');
+
+    expect([...classes]).to.have.members(['block']);
```
NullVoxPopuli: it may be worth it to have a little inline snapshot tets for these two tests -- even though the other tests don't have, because scss is not normal, the extra verbosity seems like it'd be nice

ember-scoped-css/src/build/template-plugin.test.ts
evoactivity: not sure what you mean here?

ember-scoped-css/src/build/unplugin-colocated.js
evoactivity: CWD could be different depending on how you launch the process.

eg in a monorepo`pnpm --filter vite-app build` would have CWD be the monorepo root instead of that package root, and resolve would look in the monorepo top levl node_modules.

NullVoxPopuli: i thought pnpm changed the cwd when it did that 🤔

ember-scoped-css/src/lib/css/utils.js
evoactivity: this is just testing that the postcss parse is able to get the class names it needs to rewrite, these inline snapshots would just be

```js
expect([...classes]).toMatchInlineSnapshot(`
  [
    "parent",
    "child",
  ]
`);

expect([...classes]).toMatchInlineSnapshot(`
  [
    "block",
  ]
`);
```

I'm not sure that's an improvement compared to just checking the array members we expect exist?

ember-scoped-css/src/build/template-plugin.test.ts
evoactivity: I think I've updated this to what you meant?

ember-scoped-css/src/build/unplugin-colocated.js
evoactivity: hm, maybe it does (I haven't tested that tbh), but other tools may not?

evoactivity:
Should we add code to special case BEM in sass format?
```scss
.block {
  &--modifier {
    //
  }
}
```
becomes `.block_97ew7ew`, `.block_97ew7ew--modifier` but if someone looks for `scopedClass('block--modifier')` it'll come back as `block--modifier_97ew7ew`, same in the template `<div class="block block--modifier" />` ends up as `<div class="block_97ew7ew block--modifier" />`

I hate this style of BEM class construction in scss, and since classes are scoped BEM is really not worthwhile here, so maybe not necessary for us to explicitly support?

---
Failing test if we think we should support it

```js
it('handles scoped lang="scss" BEM constructs', async () => {
  let output = await transform(`
    export const Foo = <template>
      <div class="block block--modifier">hi</div>
      <style scoped lang="scss">
        .block {
          &--modifier { color: green; }
        }
      </style>
    </template>;
  `);

  expect(templateContentsOf(output)).toMatchInlineSnapshot(`
    [
      "<div class="block_e65d154a1 block_e65d154a1--modifier">hi</div>",
    ]
  `);
});
```

ember-scoped-css/src/build/unplugin-colocated.js
NullVoxPopuli: I dunno which ones 🙈

NullVoxPopuli: but generally reducing fs access is good for macos and windows

NullVoxPopuli:
yea, I think that's fine to special case

ember-scoped-css/src/build/unplugin-colocated.js
evoactivity: I've changed it to just `const require = createRequire(config.root);` since `config.root` is defined by vite itself it should always be the correct dir where we can resolve that same version of vite.

evoactivity:
I noticed the `vite-app-layer` tests fail locally (I've fixed them), but is not included in the github CI workflow. Do you know why?

NullVoxPopuli:
probably just a mistake :( thanks for doing that!

ember-scoped-css/src/build/template-plugin.test.ts
NullVoxPopuli: this would return true for both `<style lang=scss>` (if it's still in the dom) as well as the import -- since there is ambiguity, this assertion should be sliiiiightly more specific

evoactivity:
I've added support for BEM class construction.

But, this now has me thinking on overall architecture, preproccesor code can get complicated. We could have functions outputting classes, or mixins creating other nested classes. To have full support for preprocessing I think we would need to preprocess first, and then do the class rewriting.

My idea would be to essentially replace the babel stages with zmod-ember running directly in the plugin. That should open up being able to preprocess first, put that normal CSS back into the template, and then further in the pipeline we are just rewriting classes from normal CSS, without having to worry about how those class names are created by the preprocessor.

I think what this PR provides right now would still be useful, and could be released, but before RFCing this back into ember we may want to rearchitect the pipeline to make it more robust.

ember-scoped-css/src/build/template-plugin.test.ts
```diff
@@ -28,6 +28,31 @@ async function transform(file: string, config = {}) {
...
+
+  let j = jscodeshift;
+
+  let result: string[] = [];
+
+  j(file)
+    .find(j.ImportDeclaration, {
+      source: {
+        value: (value) =>
+          typeof value === 'string' && value.includes('.ember-scoped.css?css='),
```
NullVoxPopuli: can value actually be more than a string here?

ember-scoped-css/src/build/template-plugin.test.ts
NullVoxPopuli: I see, good update!

ember-scoped-css/src/build/template-plugin.test.ts
evoactivity: on the import we generate? I don't think it could be anything but a string.

```js
import "this-is-always-a-string-isn't-it";
```

NullVoxPopuli: seems like extra complexity in this visitor thing then 🙈

NullVoxPopuli (approved):

TeriyakiBomb:
Closed PR I know, but I've heard anecdotally that sass-embedded is more performant than regular dart-sass

https://thib.me/sass-vs-sass-embedded-performance

NullVoxPopuli:
any performance improvements to this package are welcome! especially moving to rust or other performance focused tools

ember-scoped-css/src/build/unplugin-colocated.js
TeriyakiBomb: > CWD could be different depending on how you launch the process.
> 
> eg in a monorepo`pnpm --filter vite-app build` would have CWD be the monorepo root instead of that package root, and resolve would look in the monorepo top levl node_modules.

Just did the exact same thing in #387

TeriyakiBomb:
Once we sort #387, I'll put together a test for before and after perf and then if it's meaningful I'll PR

PR #387, opened 2026-03-22
Set explicit CWD for inline sass

TeriyakiBomb:
using sass mixins/functions fail silently without this

### DIFF
--- a/ember-scoped-css/src/build/unplugin-inline.js
+++ b/ember-scoped-css/src/build/unplugin-inline.js
@@ -96,7 +96,7 @@ export function inline(options = {}) {
         // Resolve Vite's preprocessCSS from the app root to ensure we find
         // the correct Vite installation (not a stale or missing one).
         try {
-          const require = createRequire(config.root);
+          const require = createRequire(path.join(config.root, 'package.json'));
           const vitePath = require.resolve('vite');
           const viteModule = await import(vitePath);
 


ember-scoped-css/src/build/unplugin-inline.js
```diff
@@ -96,7 +96,7 @@ export function inline(options = {}) {
         // Resolve Vite's preprocessCSS from the app root to ensure we find
         // the correct Vite installation (not a stale or missing one).
         try {
-          const require = createRequire(config.root);
+          const require = createRequire(path.join(config.root, 'package.json'));
```
NullVoxPopuli, line 99: any reason to not use CWD here? https://github.com/auditboard/ember-scoped-css/pull/387/changes#diff-604867ece561707648d57b1312954bc2a488263d20ed2cf37dc8a27c2c1ee685R19


Also, why does createRequire with the package.json work but without doesn't? 

can you provide the errors you're seeing and what the value of config.root is?

TeriyakiBomb: CWD works just as well tbh.

`createRequire(config.root)` returns:

```
[
  '/Users/ash/code/node_modules',
  '/Users/ash/node_modules',
  '/Users/node_modules',
  '/node_modules',
  '/Users/ash/.node_modules',
  '/Users/ash/.node_libraries',
  '/opt/homebrew/Cellar/node@24/24.14.0_1/lib/node'
]
```

With `createRequire(CWD);`, `'/Users/ash/code/ember-scoped-css/node_modules',` is at the top of the arr.

I don't get any errors beforehand, but installing into a fresh app, define a sass mixin or function and then try to use it inline and it'll fail. Compiled classes work just fine though.

NullVoxPopuli: but i mean what's the value of `config.root`? `createRequire(config.root)` should return something function like :see_no_evil: -- what you see in that array is what directories require / node's resolve look for packages in/from

TeriyakiBomb:
After a conversation with @NullVoxPopuli turns out that `createRequire` expects a filename, so adding an arbitrary file makes it work. Not sure the right answer to this to TBH + the same issue probably exists in unplugin-colocated.js

evoactivity:
I originally had `createRequire(path.join(config.root, 'package.json'));`

TeriyakiBomb:
> I originally had `createRequire(path.join(config.root, 'package.json'));`

It works but is that acceptable to you guys? It’s a little bit magic, you said somewhere else you might look to re-engineer this anyway? So is it ok for now?

NullVoxPopuli (approved):

PR #390, opened 2026-06-09
feat: pluggable CSS processors with lightningcss as default

### LINKED ISSUE #320: Migrate to lightningcss?
Is rust, so it's 100x faster: https://lightningcss.dev/

tylerturdenpants:
Closes #320.

## Summary
Pluggable CSS-processor layer. **lightningcss** is the default scoping engine; **postcss** is a configurable opt-in. Both implement the same scoping transform behind a shared interface, selected per file.

## How it works
- `src/lib/css/processors/`: `lightningcss.js` (new), `postcss.js` (existing logic moved), `index.js` resolver.
- Routing: explicit `type` wins; `.scss`/`.sass`/`.less` → postcss (lightningcss can't parse raw preprocessor syntax); everything else → lightningcss.
- `rewrite.js`/`utils.js` are thin dispatchers (the `/* file */` header + `@layer` wrapping live in the dispatcher).

## Options (vite / rollup / babel)
- `type: 'lightningcss' | 'postcss'` — force one processor for all files.
- `lightningcss: {…}` — passed through to lightningcss `transform` (e.g. `targets`, `drafts`).
- `postcss: { plugins: [] }` — user postcss plugins run before scoping.

## Transform parity
classes (single `_`), tags, `:global`, `:nth-of-type`, `@keyframes` + `animation`/`animation-name`, `@counter-style` + `list-style`/`list-style-type`, `@property` + `var()`, `@position-try` + `position-try-fallbacks`, and `@container`/`@media`/`@supports` pass-through.

## Behavior note (output normalization)
lightningcss normalizes CSS even with `minify: false` — colors (`black`→`#000`), units (`0px`→`0`), and redundant-declaration dedup. Scoped class names are unchanged. Consumers can pass `lightningcss.targets` to tune output. The `expected-dist` fixtures across the 5 test-apps were regenerated to reflect this; diffs are normalization-only.

## Performance — heads-up, this is *not* a speed win
A transform-only micro-benchmark shows **postcss is ~1.6–3.2× faster than lightningcss** for this workload:

| Workload | lightningcss | postcss |
|---|---|---|
| small (no @-rules) | 28.3 µs/op | 13.3 µs/op |
| medium, no @-rules | 129 µs/op | 78 µs/op |
| medium w/ @keyframes/@property | 272 µs/op | 83 µs/op |

Root cause: scoping is driven by JS visitor callbacks. Even a single trivial visitor makes lightningcss ~3× slower than its pure-native path (21→66 µs/op) — the Rust↔JS boundary dominates. lightningcss's well-known speed advantage applies to native-only transforms (minify/prefix), not JS-visitor-driven AST rewriting. A fast-path skips the collection pass when no referenceable at-rules are present, narrowing the common-case gap (~3.1×→~2.1×) but not closing it. In real builds the absolute cost is negligible (hundreds of µs/file). **The value of this change is features/maintenance (sourcemaps, modern-CSS via `targets`, postcss-plugin config, Vite alignment), not performance.**

## Testing
98 unit tests, lint + build clean, all 5 `test:fixture` apps and `test:ember` green.


tylerturdenpants:
Closing in favor of #391

PR #393, opened 2026-06-26
feat: scope attribute selectors

adrianbw:
## Summary

Adds scoping for CSS **attribute selectors** (`[disabled]`, `[type="text"]`, `[data-state="open"]`, …), following the same pattern used for bare element selectors: keep the attribute as the discriminator, add the generated class to the selector, and add that class to matching elements in the template.

## How it works

**CSS rewrite** (`rewrite.js`) — attribute selectors are scoped with one of two strategies (see `isRenamedClassAttribute` in `utils.js`, shared by rewrite and discovery):

| Selector | Strategy | Output |
|---|---|---|
| `[disabled]`, `[type="x"]`, `[data-*]` | generated class | `[type="x"].postfix` |
| `[class="foo"]`, `[class~="foo"]` | rename value (no generated class) | `[class~="foo_postfix"]` |
| `[class^=]`, `[class*=]` | keep value + generated class | `[class^="foo"].postfix` |
| `[class\|="foo"]`, `[class$="foo"]` | generated class + flagged by stylelint | `[class\|="foo"].postfix` |

- The generated class is added **once per compound** (`input[type="x"]` → `input.postfix[type="x"]`).
- `:global(...)` is respected; the case-insensitive `i` flag is preserved.
- Attribute names are matched **case-insensitively** (lowercased on both sides), like CSS matches HTML attributes.
- Functional pseudos (`:not/:is/:has/:where`) are rewritten as tags already are — no special-casing.

**Discovery** (`utils.js`) — collects attribute names into a new `attributes` Set; registers renamed-value class tokens in `classes`.

**Template** (`rewriteHbs.js`) — adds the generated class to elements that carry a scoped attribute, **including component invocations**: the class reaches the component's root through `...attributes` the same way the matched attribute does. Elements that spread `...attributes` are scoped whenever the file scopes by attribute at all, since any attribute may arrive at runtime (the preserved selector still decides which rules apply). Named args (`@type`) are never matched. Dynamic class values (`class={{x}}`, `class="a {{b}}"`) receive the generated class too (this also fixes a pre-existing gap for tag scoping).

**stylelint** (`no-unscoped-selectors`) — now treats attribute selectors as scoped, avoiding false positives. A new **`no-unscopable-class-attribute-selectors`** rule flags `[class|=]` (imprecise once classes are renamed) and `[class$=]` (can never match once the generated class is appended to the class attribute).

**Docs** (`css-isolation.md`) — documents attribute scoping plus a "Known limitations" section (standalone `:not()` leaks, class-attribute operator caveats, attributes that only arrive at runtime via splattributes/modifiers, and what forwarding the generated class into a component implies).

## Known limitations (documented)

- `[class$="foo"]` / `[class|="foo"]` can't be reliably scoped — flagged by the stylelint rule.
- `[class="foo"]` is an exact match and breaks if the element also receives the generated class from another selector; `[class~=]` is unaffected.
- Standalone negated selectors (`:not([disabled])`, and the pre-existing `:not(div)`) leak globally — there's no positive anchor inside the inverted `:not()`. The stylelint rule flags these.
- Attributes set by modifiers (or direct DOM mutation) aren't statically visible, so those elements aren't scoped — see the upgrade note in the docs. (`...attributes` **is** handled.)

## Possible follow-up

A build-time warning when a discovered attribute name never appears statically in the template would make the runtime-attribute case loud instead of silent. Left out of this PR to keep it reviewable.

## Tests

- `ember-scoped-css`: 112 passing (rewrite, discovery, template scoping incl. dynamic class values, `...attributes`, and case-insensitivity).
- `stylelint-ember-scoped-css`: 71 passing (attribute-as-scoped cases, `|=`/`$=` rejections, dedupe of nested-rule warnings).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


ember-scoped-css/src/lib/css/rewrite.js
```diff
@@ -33,8 +34,61 @@ function rewriteReferenceable(node, postfix) {
...
+    if (siblings[i].type === 'combinator') break;
+    if (siblings[i].type === 'class' && siblings[i].value === postfix)
+      return true;
+  }
+
+  for (let i = index + 1; i < siblings.length; i++) {
+    if (siblings[i].type === 'combinator') break;
+    if (siblings[i].type === 'class' && siblings[i].value === postfix)
+      return true;
+  }
```
adrianbw, line 57: [question](https://pullpo.io/cc?l=question) 
Is this the most-performant way to do this?

ember-scoped-css/src/lib/css/rewrite.js
```diff
@@ -50,13 +104,34 @@ function rewriteSelector(sel, postfix) {
       if (selector.type === 'class') {
         selector.value += '_' + postfix;
       } else if (selector.type === 'tag') {
-        selector.replaceWith(
-          parser.tag({ value: selector.value }),
-          parser.className({ value: postfix }),
-        );
```
adrianbw, line 56: [question](https://pullpo.io/cc?l=question) 
Just to make sure I understand: how are we reproducing the behavior of this removed block.

docs/css-isolation.md
```diff
@@ -235,3 +235,84 @@ will become
...
+class instead of adding a marker, because `.foo` is itself renamed to
+`.foo_generated`:
+
+```css
+[class~='foo'] {
+  ...;
+} /* -> [class~="foo_generated"] */
+```
+
+Only **native** elements are marked. Component invocations (`<Foo>`,
```
NullVoxPopuli: why this limitation? components can have attributes per `...attributes`, thus they can also be scoped the same as elements.

docs/css-isolation.md
```diff
@@ -235,3 +235,84 @@ will become
...
+`<foo.bar>`) are skipped so the scope marker is never forwarded into another
+component's scope. Named arguments (`@type`) and `...attributes` are not treated
```
NullVoxPopuli: > `...attributes` are not treated as attributes.

that's how all this works with components :)

docs/css-isolation.md
```diff
@@ -235,3 +235,84 @@ will become
...
+as attributes.
+
+## Known limitations
+
+These limitations stem from the marker-class model and apply to bare element
+selectors as well as attribute selectors.
+
+### Standalone negation leaks
+
+The marker class is a _positive_ signal ("this element belongs to my
```
NullVoxPopuli: this sentence contradicts the described limitation above (which is solveable anyway, and then this wouldn't be in conflict)

docs/css-isolation.md
```diff
@@ -235,3 +235,84 @@ will become
...
+the whole document:
+
+```css
+:not([disabled]) {
+} /* -> :not([disabled].generated) — leaks globally */
+:not(div) {
+} /* -> :not(div.generated)        — same leak, pre-existing for tags */
+```
+
+As long as the negation has a positive anchor in the same compound, scoping is
```
NullVoxPopuli: this is a good callout

docs/css-isolation.md
adrianbw: This is one of the assumptions that the description calls out. We can mark Component invocations, but because of splatributes, the class could end up somewhere we _really_ don't expect. If you're good with that, I'll remove.

docs/css-isolation.md
adrianbw: Help me understand the contradiction?

docs/css-isolation.md
adrianbw: I was pretty impressed that Claude found this, TBH.

ember-scoped-css/src/lib/css/rewrite.js
adrianbw: It's not on a hot path: `compoundHasMarker` only runs when we're about to add a marker, and the scan is bounded to a single compound (the nodes between combinators) — usually a handful. I kept the linear sibling scan for readability; if profiling ever showed it mattering we could track already-marked compounds in a Set instead, but that's bookkeeping for no measurable win at these sizes.

— *Automated by AI agent*

ember-scoped-css/src/lib/css/rewrite.js
adrianbw: Same output, just deferred. Tag nodes are pushed into `toMark` during the walk and the marker is inserted afterward via `addMarker` (`parent.insertAfter(node, className(postfix))`), so `div` still becomes `div.postfix`. Inserting after the walk avoids the freshly-added marker class being re-visited by the walk, and routing tags through `addMarker` lets a tag+attribute compound like `input[type="text"]` share one marker (`input.postfix[type="text"]`) instead of getting two.

— *Automated by AI agent*

docs/css-isolation.md
NullVoxPopuli: we already scope for `class`, so I think it'd be consistent

docs/css-isolation.md
NullVoxPopuli: earlier it says components aren't supported (we want them supported tho) for attribute scoping

docs/css-isolation.md
NullVoxPopuli: probably helps that this library exists in like... 5 other variations in various other ecosystems

test-apps/v2-addon-bundled/tests/rendering-test.gts
```diff
@@ -18,6 +18,12 @@ module('example', function (hooks) {
       .hasClass(
         scopedClass('green', 'v2-addon-bundled/components/from-direct/example'),
       );
+
+    // attribute selector: the button is marked via the scope class
```
NullVoxPopuli: useless comment

ember-scoped-css/src/lib/css/rewrite.js
```diff
@@ -33,8 +34,61 @@ function rewriteReferenceable(node, postfix) {
...
+  for (let i = index + 1; i < siblings.length; i++) {
+    if (siblings[i].type === 'combinator') break;
+    if (siblings[i].type === 'class' && siblings[i].value === postfix)
+      return true;
+  }
+
+  return false;
+}
+
+function addMarker(node, postfix) {
```
NullVoxPopuli: is this "marker" nomenclature new here?

I noticed it in the docs as well, and it felt out of place / new to me 🤔

ember-scoped-css/src/lib/css/rewrite.js
```diff
@@ -33,8 +34,61 @@ function rewriteReferenceable(node, postfix) {
...
+  if (compoundHasMarker(node, postfix)) return;
+
+  node.parent.insertAfter(node, parser.className({ value: postfix }));
+}
+
 function rewriteSelector(sel, postfix) {
   const transform = (selectors) => {
+    // Nodes we need to mark with the scope class. We collect them during the
```
NullVoxPopuli: question for the docs:
- what is a marker class
- what determines if we "mark" a node?
- what is marking?

ember-scoped-css/src/lib/css/rewrite.test.ts
```diff
@@ -2,6 +2,14 @@ import { describe, expect, it } from 'vitest';
 
 import { rewriteCss } from './rewrite.js';
 
+/** Rewrite a single rule and return just the scoped selector. */
+function scopeSelector(css) {
+  return rewriteCss(css, 'postfix', 'foo.css')
```
NullVoxPopuli: we should not use regex here, and this has gotten complicated enough now where I think we should pull in the selector parser, and use that. replacing and regexes are annoying to deal with, and are too specific.

Additionally, since this file is massive, all these new tests should probably be their own test file.

rewrite.attributes.test.ts

ember-scoped-css/src/lib/css/utils.js
```diff
@@ -97,13 +98,31 @@ function resolveNestedSassSelector(node) {
...
         tags.add(selector.value);
+      } else if (selector.type === 'attribute') {
+        if (
+          selector.attribute === 'class' &&
+          (selector.operator === '=' || selector.operator === '~=') &&
+          selector.value
+        ) {
+          // Bucket A: the value names real class(es), which get renamed in the
+          // template. Register them as classes so the renaming happens.
+          for (let token of selector.value.split(/\s+/).filter(Boolean)) {
```
NullVoxPopuli: does 'postcss-selector-parser' really not give us anything better here?

ember-scoped-css/src/lib/css/utils.js
```diff
@@ -125,6 +144,36 @@ if (import.meta.vitest) {
     expect([...tags]).to.have.members(['div']);
   });
 
+  it('collects attribute names for marker-based scoping', function () {
+    const css = '[disabled] [data-x="y"] { color: red; }';
+    const { attributes } = getCSSContentInfo(css);
+
+    expect([...attributes]).to.have.members(['disabled', 'data-x']);
```
NullVoxPopuli: why are we creating new arrays for no reason?

ember-scoped-css/src/lib/css/utils.js
```diff
@@ -125,6 +144,36 @@ if (import.meta.vitest) {
...
+  });
+
+  it('treats =/~= class attribute values as renamed classes', function () {
+    const css = '[class="foo bar"] [class~="baz"] { color: red; }';
+    const { classes, attributes } = getCSSContentInfo(css);
+
+    expect([...classes]).to.have.members(['foo', 'bar', 'baz']);
```
NullVoxPopuli: why are we creating new arrays for no reason? 

(and the rest of these tests)

ember-scoped-css/src/lib/rewriteHbs.js
```diff
@@ -2,7 +2,28 @@ import * as recast from 'ember-template-recast';
...
+ * carry an unknown set of attributes and `@args` are component arguments, so
+ * neither counts as a matchable attribute.
+ */
+function elementHasScopedAttribute(node, attributes) {
+  if (attributes.size === 0) return false;
+
+  return node.attributes.some((attr) => {
+    let name = attr.name;
+
+    if (name === '...attributes' || name.startsWith('@')) return false;
```
NullVoxPopuli: is this line needed? seems like since we are in control of what goes in to the attributes set, we could just... not put `...attributes` or `@args` in it, and then make this function either a one liner

function feels overly defensive

ember-scoped-css/src/lib/rewriteHbs.test.ts
NullVoxPopuli, line 1: this file is about attribute rewriting, so it should be named as such


also, all the questions about "marking" remain. has that word been used before this PR?

NullVoxPopuli (changes requested):
some todos for me as I take over the PR

NullVoxPopuli-ai-agent (commented):
Review of the attribute-selector scoping feature. The direction is right and consistent with the library's purpose — attribute selectors leaking globally was itself a violation of the scoping promise, and this PR closes it with the same postfix-class/rename machinery used for tags and classes. The recurring theme in the findings is not the tightening but **silent divergence between the two halves**: the CSS rewrite now unconditionally requires the postfix class, so any path where the template side can't produce one yields styles that neither leak nor apply, with no signal to the author at build, lint, or runtime. The findings in the inline comments are all instances of this, each verified by running the branch's actual `rewriteCss`/`rewriteHbs` pipeline. The asks are about making those cases work or making them loud (fix the emitter where possible, otherwise warn/lint/document) — not about preserving any old leaking behavior.

**Terminology used in the comments below**, since neither term is defined in the repo: **Bucket A** / **Bucket B** are this PR's comment shorthand — Bucket A = class-targeting attribute selectors whose value names real classes (`[class="foo"]`, `[class~="foo"]`), scoped by renaming the value; Bucket B = every other attribute selector, scoped by appending the per-file **postfix class** (the code's term; the docs call it the "generated class"). Earlier revisions of these comments said "marker" for the postfix class — that word comes from the PR description and appears nowhere in the repo, so I've standardized on the code's term.

Inline comments cover: `[class$=]` being unconditionally unsatisfiable (not just the documented edge case), dynamic class bindings (`class={{...}}`) never receiving the postfix class, `[class="foo"]` exact-match breaking when combined with any tag/Bucket B selector, the shared postfix class forwarding more scope than the matched selector on component invocations, the `...attributes`/modifier upgrade gap, attribute-name case sensitivity, duplicate stylelint warnings, and the duplicated Bucket A/B predicate — plus nomenclature notes.

Smaller notes not worth their own threads:

- The `walkRules` skeleton in the new stylelint rule (~35 lines: at-rule skip, nesting-parent skip, flatten, display, report) is copy-pasted from `no-unscoped-selectors`, along with a third copy of `isInsideGlobal`. Extracting a shared traversal helper would keep the two rules from drifting. A cheap whole-file `|=` pre-check would also skip per-rule flattening work that now runs over every file in the default config for an operator that almost never occurs.
- `compoundHasPostfixClass`'s rightward scan is unreachable: insertions happen post-walk in document order, each immediately after its node, so an existing postfix class in the compound is always to the left. Harmless, but the "walk left and right" comment misdescribes the invariant.
- Deleting the `@media`/`@container`/nesting reject tests from `no-unscoped-selectors` leaves the at-rule flattening path with no coverage in either direction. Re-adding a couple as accepts (`[attr]`) and rejects (`#id`) would restore it.
- `rewriteHbs`'s new defaulted 5th positional parameter means a caller that forgets to thread `attributes` runs clean with attribute scoping silently off; an options object (matching `templatePlugin`'s shape) would make the omission visible at the call site.
- `elementHasScopedAttribute` runs for every element in every template even when the attributes set is empty (the common case). A `attributes.size > 0 &&` guard restores the pre-PR fast path.
- The PR description looks stale in two places: it says component invocations are *skipped* (the code, docs, and tests all mark them — see the open-question comment inline) and that `[class|=]` emits a `console.warn` (the stylelint rule replaced it).
- Terminology: the code term for the generated class is "postfix" (`renameClass.js`, `path/utils.js`) and most of the PR follows it, but `template-plugin.js` coins "the scope class", the docs say "generated class", a test comment says "generated scoping class", and the PR description says "marker class". Grep-ability suffers — worth picking "postfix class" in code and "generated class" in user-facing docs, consistently.

A thought on a deeper fix for the recurring theme: a build-time warning when a discovered attribute name never appears statically in any element of the template would catch most of these silent-loss cases at once — keeping the scoping while making the misses loud.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

ember-scoped-css/src/lib/css/rewrite.js
```diff
@@ -50,13 +104,32 @@ function rewriteSelector(sel, postfix) {
...
+        } else {
+          // Bucket B: keep the discriminator, scope with the postfix class.
+          //
+          // `[class|="foo"]` cannot be fully scoped via renaming and only
+          // matches by the preserved value, so it falls back to the postfix
+          // class like any other Bucket B selector. That imprecision is
+          // surfaced to authors by the
+          // `ember-scoped-css/no-unscopable-class-attribute-selector`
+          // stylelint rule rather than a runtime warning.
+          needsPostfixClass.push(selector);
```
NullVoxPopuli-ai-agent, line 103: **`[class$="foo"]` becomes unconditionally unsatisfiable, not just the documented edge case.**

*(Terminology, borrowing this PR's comment shorthand: **Bucket A** = class-targeting attribute selectors whose value names real classes — `[class="foo"]`/`[class~="foo"]` — scoped by renaming the value the way `.foo` → `.foo_pfx` is renamed; **Bucket B** = every other attribute selector, scoped by appending the **postfix class**, the per-file generated class the docs call the "generated class".)*

Bucket B puts `class` into the `attributes` set, so every element that has a `class` attribute gets the postfix class — and `rewriteHbs` always appends it *last* (`classAttr.value.chars += ' ' + postfix`). So the rendered class string always ends with the 9-char hex postfix, while the rewritten selector `[class$="foo"].e1234abcd` simultaneously requires the postfix class *and* a class string ending in `"foo"`. Those are mutually exclusive (the `e`+md5-hex postfix can never end in `foo` — `o` isn't a hex digit).

Verified end-to-end on this branch: CSS `[class$="foo"] {}` + `<div class="foo">` → `<div class="foo e1234abcd">` + `[class$="foo"].e1234abcd` → matches nothing, ever (for static templates; only runtime `classList.add` after the postfix class could re-satisfy it).

docs/css-isolation.md:325 says `$=` only breaks "if the matched class is also renamed through its own `.foo` selector" — that's not the actual failure mode. Given it can never match, consider routing `$=` to the same stylelint rule as `|=` (or documenting it as fully unsupported) rather than presenting it as scoped-but-fragile.

ember-scoped-css/src/lib/rewriteHbs.js
```diff
@@ -46,18 +61,24 @@ export function templatePlugin({ classes, tags, postfix }) {
...
+      const shouldScope =
+        tags.has(node.tag) || elementHasScopedAttribute(node, attributes);
+
+      if (!shouldScope) return;
+
+      // check if class attribute already exists
+      const classAttr = node.attributes.find((attr) => attr.name === 'class');
+
+      if (classAttr) {
+        classAttr.value.chars += ' ' + postfix;
```
NullVoxPopuli-ai-agent: **Elements with dynamic class bindings never receive the postfix class, so their attribute-selector styles silently die.**

`classAttr.value.chars` only exists when the value is a TextNode. For `class={{this.cls}}` (MustacheStatement) or `class="foo {{bar}}"` (ConcatStatement), this `+=` writes a bogus `chars` property that ember-template-recast ignores when printing — verified empirically on this branch:

- `<input type="text" class={{this.cls}}>` with `attributes = {'type'}` → output unchanged, no postfix class added
- `<input type="text" class="foo {{bar}}">` → output unchanged, no postfix class added

Meanwhile the CSS was rewritten to `[type="text"].pfx`, so a rule that matched before this PR (by leaking) now silently applies to nothing.

The `.chars +=` bug pre-exists for tag selectors, but this PR widens its trigger surface to every element carrying a scoped attribute name — dynamic class bindings are common enough that this seems worth fixing at the source (handle ConcatStatement/MustacheStatement by wrapping in a ConcatStatement) rather than shipping around.

ember-scoped-css/src/lib/css/rewrite.js
```diff
@@ -50,13 +104,32 @@ function rewriteSelector(sel, postfix) {
...
-          parser.tag({ value: selector.value }),
-          parser.className({ value: postfix }),
-        );
+        needsPostfixClass.push(selector);
+      } else if (selector.type === 'attribute') {
+        if (isRenamedClassAttribute(selector)) {
+          // Bucket A: rewrite the value to track the renamed class. The
+          // renamed token (e.g. `foo_postfix`) is already unique per file, so
+          // no postfix class is needed.
+          selector.value = renameClass(selector.value, postfix);
```
NullVoxPopuli-ai-agent, line 95: **Bucket A exact-match `[class="foo"]` breaks whenever the same element is also scoped by any tag or Bucket B selector in the same file.**

*(Terminology, borrowing this PR's comment shorthand: **Bucket A** = class-targeting attribute selectors whose value names real classes — `[class="foo"]`/`[class~="foo"]` — scoped by renaming the value the way `.foo` → `.foo_pfx` is renamed; **Bucket B** = every other attribute selector, scoped by appending the **postfix class**, the per-file generated class the docs call the "generated class".)*

The rename keeps the `=` operator, but `rewriteHbs` appends the standalone postfix class to any element that also matches a tag selector or carries a Bucket B attribute. Verified on this branch:

- CSS `div {} [class="foo"] {}` + `<div class="foo">` → element `class="foo_pfx pfx"`, selector `[class="foo_pfx"]` — exact string match fails, rule never applies
- Same within one compound: `[class="foo"][disabled]` → `[class="foo_pfx"][disabled].pfx` vs `<button class="foo_pfx pfx" disabled>` — dead

`~=` is safe (token match); only `=` has this collision. The docs present both as reliably scoped. Either add `=` to the caveats, rewrite `=` to `~=`-style semantics where possible, or treat it like the other unreliable operators.

ember-scoped-css/src/lib/rewriteHbs.js
```diff
@@ -46,18 +61,24 @@ export function templatePlugin({ classes, tags, postfix }) {
...
-          // push class attribute
-          node.attributes.push(
-            recast.builders.attr('class', recast.builders.text(postfix)),
-          );
-        }
+      // An element is in scope if its tag matches a tag selector, or if it
+      // carries an attribute named in a scoped attribute selector. We add the
+      // postfix class at most once regardless of how many things matched.
+      const shouldScope =
+        tags.has(node.tag) || elementHasScopedAttribute(node, attributes);
```
NullVoxPopuli-ai-agent, line 81: **Re: the open question — the issue isn't marking invocations per se, it's that the shared per-file postfix class grants more than the matched selector.**

*(Terminology: the **postfix class** is the per-file generated class — the docs' "generated class". "Bucket B" below is this PR's comment shorthand for attribute selectors scoped by appending it, i.e. everything except `[class=]`/`[class~=]`, which are "Bucket A" and scoped by renaming the value.)*

Forwarding scoped styles through `...attributes` is consistent with how the library already works for classes: `<Child class="foo" />` forwards `foo_pfx`, and exactly the parent's `.foo` rules apply to the child's root — precise, because the renamed class encodes *which* selector matched. The new attribute scoping forwards the *generic* postfix class instead, and that class means "this element is in scope for every selector in the file that was scoped by appending it". Two compounding effects, both verified on this branch:

1. Elements are matched by attribute *name* only, so `<Child data-variant="secondary" />` gets the postfix class even though the CSS only has `[data-variant="primary"]`.
2. Once the postfix class reaches the child's root, that element matches **all** of the parent's postfix-scoped selectors — tag rules (`button {}` → `button.pfx`) and Bucket B rules — not just the attribute rule that caused it to be added. Concretely: parent CSS `[data-variant="primary"] { color: A } button { color: B }`, parent template `<Child data-variant="primary" />`, child template `<button ...attributes>` → the child's root button gets `color: B` from the parent's plain `button` rule, which previously could never cross the component boundary.

Options as I see them: (a) keep invocation marking as the scoped-forwarding feature it is, and document that carrying a scoped attribute name opts the child's root into the parent's tag/attribute rules wholesale; (b) restrict it to native elements — strictest isolation, but gives up the genuinely useful `[data-variant]` → `<Child>` forwarding your test pins down; or (c) longer-term, make attribute scoping precise the way class renaming already is (e.g. selector-specific classes), which fixes both the value-insensitivity and the tag-rule crossover. Given the library's goal is scoping styles rather than withholding them, (a) now with (c) as the eventual fix seems most in keeping — but whichever way you land, the current behavior deserves an explicit line in the docs, and the PR description (which says invocations are *skipped*) no longer matches the code/docs/tests.

docs/css-isolation.md
```diff
@@ -235,3 +235,93 @@ will become
...
+becomes
+
+```css
+/* components/first.css */
+[class~='foo_generated-first'] {
+  ...;
+}
+```
+
+Component invocations are scoped too. A component receives the generated class through its `...attributes` the same way it receives the matched attribute, so `[type='text']` scopes `<Foo type="text" />` just as it scopes `<input type="text" />`. Named arguments such as `@type` are not HTML attributes, and `...attributes` carries an unknown set of attributes, so neither is matched.
```
NullVoxPopuli-ai-agent, line 289: **The tightening here is right — it just needs an upgrade note and ideally a build-time signal.**

To be clear about direction: the old behavior was a leak — `[disabled]` in a component's CSS also styled every other component's disabled elements, contrary to the library's core promise — so scoping it is correct, and restoring the leak is not the ask. The concern is that one common pattern crosses from "leaks but appears to work" to "silently applies to nothing" with no signal at any layer:

A component whose own CSS says `[disabled] { ... }` with template `<button ...attributes>` — the attribute arrives from callers via splattributes (or from a modifier) — now compiles to `[disabled].pfx`, but the button never receives the generated class: `...attributes` parses as an attr literally named `...attributes`, and modifier-set attributes aren't visible statically at all. And since `no-unscoped-selectors` now *accepts* `[disabled]` as scoped, lint stays quiet too.

This sentence does document the receiving-side gap, but it's framed around invocation-side scoping, element modifiers aren't mentioned, and nothing flags this as a behavior change for existing code. Suggest: an explicit "upgrading" note in Known limitations, plus a build-time warning when a discovered attribute name never appears statically in the template — that would make this whole class of silent misses loud, the same way the stylelint rules already surface the other unscopable cases.

ember-scoped-css/src/lib/css/utils.js
```diff
@@ -97,13 +98,37 @@ function resolveNestedSassSelector(node) {
...
+          // space-separated class names is on us (same tokenization as
+          // `renameClass`).
+          for (let token of selector.value.split(/\s+/).filter(Boolean)) {
+            classes.add(token);
+          }
+        } else {
+          // Bucket B: elements carrying this attribute get the postfix class.
+          // For class-target operators (^=, *=, $=, |=) and presence, the
+          // name is `class`.
+          attributes.add(selector.attribute);
```
NullVoxPopuli-ai-agent: **Attribute-name matching is case-sensitive here, but CSS matches HTML attribute names ASCII case-insensitively.**

`[TYPE="submit"]` matches `<button type="submit">` in the browser (and did before this PR, since attribute selectors were left untouched). Now `'TYPE'` goes into the set verbatim, `elementHasScopedAttribute` compares against the Glimmer attr name `'type'`, no postfix class is added to the element, and the rewritten `[TYPE="submit"].pfx` matches nothing. The reverse direction (`<button TYPE="submit">` vs `[type="submit"]`) fails the same way.

Admittedly rare, and parallel to the pre-existing tag-path behavior (`DIV {}` vs `<div>`), but lowercasing on both sides would close it cheaply — with care for case-sensitive-by-spec SVG/foreign attributes like `viewBox` before blanket-lowercasing.

stylelint-ember-scoped-css/src/rules/no-unscopable-class-attribute-selectors/index.js
```diff
@@ -0,0 +1,110 @@
...
+      const hasChildRules = rule.nodes.some((node) => node.type === 'rule');
+
+      if (!hasDirectDecls && hasChildRules) {
+        return;
+      }
+
+      const flattenedSelectors = flattenNestedSelectorsForRule(rule, result);
+
+      for (const flattened of flattenedSelectors) {
+        flattened.resolvedSelectors.each((selectorNode) => {
```
NullVoxPopuli-ai-agent, line 73: **Duplicate warnings for nested rules under multi-selector parents.**

Verified: `.a, .b { [class|="foo"] {} }` emits two byte-identical warnings. `flattenNestedSelectorsForRule` returns one entry per *resolved* parent selector, each carrying the same unresolved inner selector, and both pass `hasUnscopableClassAttribute` with the report anchored to the same node/message — `stylelint.utils.report` doesn't dedupe.

The sibling `no-unscoped-selectors` has the same pattern, but there it only duplicates when the parents are themselves unscoped; here it fires under perfectly normal scoped parents, so it's more user-visible. Since the violation belongs to the inner rule itself (not the resolution), deduping by the unresolved `flattened.selector` — or just breaking after the first report per rule — would fix it.

ember-scoped-css/src/lib/css/utils.js
```diff
@@ -97,13 +98,37 @@ function resolveNestedSassSelector(node) {
...
+
+      if (selector.type === 'class') {
         classes.add(selector.value);
-      } else if (selector.type === 'tag' && !isInsideGlobal(selector)) {
+      } else if (selector.type === 'tag') {
         tags.add(selector.value);
+      } else if (selector.type === 'attribute') {
+        if (
+          selector.attribute === 'class' &&
+          (selector.operator === '=' || selector.operator === '~=') &&
```
NullVoxPopuli-ai-agent: **The Bucket A/B classification now exists as two parallel copies that must stay bit-for-bit identical.**

*(Terminology, borrowing this PR's comment shorthand: **Bucket A** = class-targeting attribute selectors whose value names real classes — `[class="foo"]`/`[class~="foo"]` — scoped by renaming the value the way `.foo` → `.foo_pfx` is renamed; **Bucket B** = every other attribute selector, scoped by appending the **postfix class**, the per-file generated class the docs call the "generated class".)*

This inline condition duplicates `isRenamedClassAttribute` in rewrite.js. Discovery (here) decides which class tokens the *template* renames; rewrite.js decides which attribute values the *CSS* renames. If a future change touches one copy only (say, adding an operator or tightening the value check), templates and CSS classify the same selector into different buckets and the styles die silently — and each side's tests only validate its own copy, so nothing catches the drift.

Suggest exporting one shared predicate (rewrite.js already imports from this area). Same story for the value tokenization: the `split(/\s+/).filter(Boolean)` below duplicates `renameClass`'s tokenization, as its own comment admits — a shared `splitClassList` helper in renameClass.js would pin them together.

ember-scoped-css/src/lib/css/rewrite.js
```diff
@@ -33,8 +34,61 @@ function rewriteReferenceable(node, postfix) {
...
+/**
+ * Class-targeting attribute selectors whose value is a real class name
+ * (`[class="foo"]`, `[class~="foo"]`) are scoped by renaming the value the same
+ * way `.foo` is renamed to `.foo_postfix`. The renamed token is unique, so the
+ * postfix class is not needed (mirrors the class-renaming strategy).
+ *
+ * Every other attribute selector keeps its discriminator and is scoped by
+ * adding the postfix class (mirrors the tag strategy: `div` -> `div.postfix`).
+ */
+function isRenamedClassAttribute(node) {
```
NullVoxPopuli-ai-agent: Nomenclature: "Bucket A" / "Bucket B" are used in comments here, in utils.js, and in six test titles, but defined nowhere in the repo — the only definition lives in the PR description, which future maintainers won't have at hand when reading a failing test titled "(Bucket A, no postfix class)".

For readers of this thread, the definitions from the PR description: **Bucket A** = class-targeting attribute selectors whose value names real classes (`[class="foo"]`, `[class~="foo"]`), scoped by renaming the value; **Bucket B** = every other attribute selector, scoped by appending the generated/postfix class.

The docs describe the same split without the bucket terms. Either define the terms once in a comment that both rewrite.js and utils.js point to, or use descriptive names (e.g. renamed-value vs postfix-scoped) so the code is self-explaining.

ember-scoped-css/src/lib/css/utils.js
```diff
@@ -97,13 +98,37 @@ function resolveNestedSassSelector(node) {
   return selector.replace(/&/g, resolvedParent);
 }
 
-function getClassesAndTags(sel, classes, tags) {
+function getClassesAndTags(sel, classes, tags, attributes) {
```
NullVoxPopuli-ai-agent: Nomenclature: `getClassesAndTags` no longer describes what it does — it now also collects attribute names and registers `[class=...]` values as classes. Someone grepping for where scoped attributes are collected won't find anything named for it. A neutral name like `collectSelectorInfo` would match the file's existing `...Info` convention (`getCSSInfo`, `getCSSContentInfo`).

stylelint-ember-scoped-css/src/rules/no-unscopable-class-attribute-selector/index.js
```diff
@@ -0,0 +1,110 @@
+import stylelint from 'stylelint';
+import flattenNestedSelectorsForRule from 'stylelint/lib/utils/flattenNestedSelectorsForRule.mjs';
+
+import buildDisplaySelector from '../../utils/buildDisplaySelector.js';
+import { AT_RULES_WITH_STYLE_RULES } from '../no-unscoped-selectors/index.js';
+
+const ruleName = 'ember-scoped-css/no-unscopable-class-attribute-selector';
```
NullVoxPopuli-ai-agent: Naming, worth settling before release since rule names are public API: the plugin's only other rule is plural (`no-unscoped-selectors`) while this one is singular, so users will config the wrong form from memory. Also "unscopable" contradicts the rule's own message — which says the selector *is* scoped by the generated class, just imprecisely. Something like `no-unscopable-class-attribute-selectors` (or folding the check into `no-unscoped-selectors`, which would also spare the duplicated traversal skeleton) would avoid both.

docs/css-isolation.md
```diff
@@ -288,6 +288,11 @@ becomes
 
 Component invocations are scoped too. A component receives the generated class through its `...attributes` the same way it receives the matched attribute, so `[type='text']` scopes `<Foo type="text" />` just as it scopes `<input type="text" />`. Named arguments such as `@type` are not HTML attributes, and `...attributes` carries an unknown set of attributes, so neither is matched.
 
+Two things to be aware of with this forwarding:
+
+- Elements are matched by attribute **name**, not value: `<Foo data-variant="secondary" />` receives the generated class even if the CSS only has `[data-variant="primary"]` (the preserved attribute selector still decides which rule actually applies).
```
NullVoxPopuli: This is unclear. please add code examples

ember-scoped-css/src/lib/css/rewrite.js
```diff
@@ -106,20 +89,20 @@ function rewriteSelector(sel, postfix) {
       } else if (selector.type === 'tag') {
         needsPostfixClass.push(selector);
       } else if (selector.type === 'attribute') {
+        // See `isRenamedClassAttribute` for the two scoping strategies.
```
NullVoxPopuli: this comment is pointless

ember-scoped-css/src/lib/css/rewrite.js
```diff
@@ -106,20 +89,20 @@ function rewriteSelector(sel, postfix) {
...
         if (isRenamedClassAttribute(selector)) {
-          // Bucket A: rewrite the value to track the renamed class. The
-          // renamed token (e.g. `foo_postfix`) is already unique per file, so
-          // no postfix class is needed.
+          // Renamed-value strategy: the renamed token (e.g. `foo_postfix`) is
```
NullVoxPopuli: you don't need to prefix/name your comments. the text left of the colon is dumb. remove it

ember-scoped-css/src/lib/css/rewrite.js
```diff
@@ -106,20 +89,20 @@ function rewriteSelector(sel, postfix) {
...
+          // already unique per file, so no postfix class is needed.
           selector.value = renameClass(selector.value, postfix);
           if (!selector.quoteMark) selector.quoteMark = '"';
         } else {
-          // Bucket B: keep the discriminator, scope with the postfix class.
+          // Postfix-class strategy: keep the discriminator, scope with the
```
NullVoxPopuli: same here

docs/css-isolation.md
```diff
@@ -288,10 +288,62 @@ becomes
 
 Component invocations are scoped too. A component receives the generated class through its `...attributes` the same way it receives the matched attribute, so `[type='text']` scopes `<Foo type="text" />` just as it scopes `<input type="text" />`. Named arguments such as `@type` are not HTML attributes, and `...attributes` carries an unknown set of attributes, so neither is matched.
 
-Two things to be aware of with this forwarding:
+Two things to be aware of with this forwarding.
 
-- Elements are matched by attribute **name**, not value: `<Foo data-variant="secondary" />` receives the generated class even if the CSS only has `[data-variant="primary"]` (the preserved attribute selector still decides which rule actually applies).
-- The generated class is shared by every selector in the file that is scoped with it. Once it reaches the child's root element through `...attributes`, that element matches **all** of the file's generated-class rules — including bare tag rules like `button { }` — not just the attribute rule that caused it to be forwarded.
+First, elements are matched by attribute **name**, not value. Given this CSS:
```
NullVoxPopuli: if we're going to have code examples of input code, we should also show what the output would be

docs/css-isolation.md
```diff
@@ -235,3 +235,191 @@ will become
...
+[type='text'] {
+  ...;
+}
+```
+
+Output:
+
+```html
+<!-- components/first.hbs -->
+<input type="text" class="generated-first" />
```
NullVoxPopuli: why does this say "generated-first"?

it should probably just be a hash or something that doesn't carry semantic meaning so as to not accidentally give readers a chance to derive meaning from where there is actually none

docs/css-isolation.md
```diff
@@ -235,3 +235,191 @@ will become
...
+
+Both rules now match every element that isn't a disabled (or `div`) element in the current component, including elements in other components and the rest of the page. As long as the negation has a positive anchor in the same compound the scope is preserved, so `button:not([disabled])` becomes `button.e55555f9d:not([disabled].e55555f9d)` and stays scoped. The `ember-scoped-css/no-unscoped-selectors` stylelint rule flags standalone negations so this is caught before it ships.
+
+### Attributes that only appear at runtime
+
+Scoping looks at the attributes written literally in the template. An attribute that arrives some other way — through `...attributes` from a caller, or set by a modifier — is not visible at build time, so the element does not receive the generated class and the rule matches nothing:
+
+```css
+/* components/first.css */
+[disabled] {
```
NullVoxPopuli, line 410: what's the output of this CSS file?

docs/css-isolation.md
```diff
@@ -235,3 +235,208 @@ will become
...
+```css
+/* components/first.css */
+[disabled].e55555f9d {
+  ...;
+}
+```
+
+```html
+<!-- components/first.hbs -->
+<button ...attributes>...</button>
```
NullVoxPopuli: I think this is actually a bug, in this case, the element with `...attributes` should have the class/postfix added to it

NullVoxPopuli (approved):

PR #397, opened 2026-07-19
feat: support additional roots across monorepo packages

alexraputa:


test-apps/vite-app-workspace/app/package.json
```diff
@@ -0,0 +1,30 @@
+{
+  "name": "vite-app-workspace-app",
```
NullVoxPopuli:
```suggestion
  "name": "vite-app-workspace-library",
```

test-apps/vite-app-monorepo/test-app/tests/monorepo-dependency-test.gjs
NullVoxPopuli, line 1: thanks for the test!

test-apps/vite-app-monorepo/test-app/src/additional-other/prefix-collision-card.gjs
alexraputa, line 1: It verifies that the configured additional root matches directory boundaries and doesn’t accidentally include paths that only share the same prefix.

I couldn’t find another test covering this case.

NullVoxPopuli (approved):

PR #401, opened 2026-08-05
Rewrite class names in unquoted class attributes

adrianbw:
## Problem

Glimmer parses a **quoted** attribute value as a `ConcatStatement` and an **unquoted** one as a bare `MustacheStatement`. The `AttrNode` visitor in `rewriteHbs.js` handled `TextNode` and `ConcatStatement` only, so the unquoted form fell through untouched:

```hbs
class="{{if x 'a' 'b'}}"   {{!-- rewritten to a_HASH / b_HASH --}}
class={{if x 'a' 'b'}}     {{!-- untouched, so the styles silently never apply --}}
```

Nothing errors and nothing warns — the class name just stays unscoped and the rule never matches. That silence is what makes it worth fixing.

## Fix

Three parts. Parts 2 and 3 came out of review, and both correct behaviour that predates this PR on the quoted path.

**1. Handle the unquoted form.** A bare `MustacheStatement` attribute value gets the same treatment as the mustache parts of a `ConcatStatement`.

**2. Rename only literals that reach the class attribute.** The original traversal renamed *every* `StringLiteral` in the mustache, including ones that are data rather than class names:

```hbs
{{if (checkAlphabet "a" "b") "a" "b"}}
{{!-- renaming everything: --}} {{if (checkAlphabet "a_HASH" "b_HASH") "a_HASH" "b_HASH"}}
{{!-- this PR:             --}} {{if (checkAlphabet "a" "b") "a_HASH" "b_HASH"}}
```

`{{if (eq this.mode "a") "a" "b"}}` is the same defect with teeth: postfixing the comparand stops the comparison ever matching, selecting the wrong branch at runtime. So the walk descends from the attribute value and propagates only through helpers whose result is assembled from their own params:

| slot | position | descend? |
| --- | --- | --- |
| the mustache's own value | class name | yes |
| `concat` params | class name, subject to part 3 | yes |
| `if`/`unless` params 1..n — the branches | class name (one becomes the value) | yes |
| `if`/`unless` param 0 — the condition | data (it becomes a boolean) | **no** |
| any param of an unrecognised helper | data (contract unknown) | **no** |

**3. Check the name is really that helper, and respect `concat`'s boundaries.**

*Shadowing.* A block param wins over both a helper and a keyword, and a shadowed name is syntactically identical to the real one — `{{#let x as |concat|}}` puts something else entirely behind it. Block params are now tracked and shadowed names treated as opaque. Element block params (`<Foo as |concat|>`) needed separate tracking: the `All` visitor only runs for node types that have no visitor of their own, so `ElementNode` never reached the ancestor stack.

*Fusing.* `concat` joins its params into one string, so a literal is only a class name of its own when a boundary separates it from its neighbours:

```hbs
{{concat "a" "-suffix"}}   {{!-- ONE class, a-suffix --}}
```

Postfixing the `"a"` emitted `a_HASH-suffix` while the CSS rewrite turned `.a-suffix` into `.a-suffix_HASH` — the postfix ended up buried mid-name and matched nothing. Literal params are folded before renaming (`a-suffix` -> `a-suffix_HASH`), and where a param is dynamic only the literals a boundary separates from it are renamed (`{{concat "a " this.x}}` yes, `{{concat "a" this.x}}` no).

### Known limitations

- A literal passed to an **unrecognised** helper (`{{myClassHelper "a"}}`) is not renamed. `{{scopedClass "a"}}` is the sanctioned way to rename a literal, so there is an explicit opt-in. This is a false negative — a class quietly not renamed — rather than corruption, and it is pinned by a test.
- Matching is **by name**. A strict-mode template has to import `concat`, and the module it came from is not visible from the template, so a `concat` imported from elsewhere is taken at face value. An earlier revision of this PR distrusted `concat` in strict mode instead; that was reverted because it removed renaming that already works today (the quoted `class="{{concat 'a' ' ' 'b'}}"` path is handled in `.gjs` on `main`), and a regression against shipped behaviour is a worse trade than the residual risk. What remains is `main`'s exposure minus the shadowing hole.
- `{{concat "a" (if C "-on" "-off")}}` is left alone, though its value set (`a-on`, `a-off`) is statically enumerable and could in principle be distributed into `{{if C "a-on_HASH" "a-off_HASH"}}`. Deliberately out of scope here; it is a new capability rather than a correctness fix, and it needs a cap on combinatorial blowup.

## Behaviour

With `classes = {a, b}` and `global-thing` deliberately absent from the sibling CSS:

| input | output |
| --- | --- |
| `class={{if x "a" "b"}}` | `class={{if x "a_HASH" "b_HASH"}}` |
| `class={{if x "global-thing"}}` | unchanged |
| `class={{if x "a" "global-thing"}}` | only `"a"` rewritten |
| `class={{concat "a" " " "b"}}` | both rewritten |
| `class={{this.fooClass}}` | unchanged |
| `class="a {{if x 'b'}}"` | unchanged from today |
| `class={{if (eq this.mode "a") "a" "b"}}` | only the branches rewritten |
| `class={{someHelper "a"}}` | unchanged |
| `class={{concat "a" "-suffix"}}` | folded, then rewritten as one class |
| `{{#let x as \|concat\|}}…{{concat "a" " " "b"}}` | unchanged (shadowed) |

Three properties make this safe, all covered explicitly: **globals are left alone** (the `classesInCss` guard), **data is left alone** (position in the expression tree), and **shadowed names are left alone**.

## Testing

- `rewriteHbs.classes.test.ts` — 36 cases across the rows above plus regressions: already-postfixed literals aren't double-postfixed, `data-x=` and `@class=` untouched, `unless` and two-param `if` behave like `if`, block-param scope does not leak past its block, and elements scoped by tag still get the postfix class.
- Each round's tests were run against the previous commit of this branch and fail there.
- **Browser e2e, chosen so the bug changes rendered style rather than just the emitted string:**
  - `vite-app` (strict mode): `{{if (eq mode "chosen") "chosen" "other"}}` — without the fix the comparand is postfixed, the comparison fails, and the wrong branch renders. Plus an imported `{{concat}}`, covering the case the reverted strict-mode revision would have regressed.
  - `vite-app-with-compat` (loose mode): `{{concat "fused" "-part"}}` against a real `.fused-part` rule. Confirmed failing without the fix.
- **Full CI test-app matrix green** — all 6 apps, dev **and** prod builds. This matters because parts 2 and 3 change the pre-existing quoted path. `vite-app-with-compat`'s `{{if @isTrue (scopedClass "is-true") (scopedClass "is-false")}}` still works: `scopedClass` subexpressions in branch slots are handled by the dedicated `SubExpression` visitor.
- Full addon suite green (147 tests). `pnpm lint` green for the addon and both test-apps.

That PR adds an ESLint rule (`auditboard/no-unrewritable-scoped-class`) covering these gaps from the consumer side.

- Its **JS-computed case remains necessary** — a class name computed in JS still can't be resolved at build time.
- Its **unquoted-literal case is redundant only where this PR rewrites correctly**. For a literal handed to an unrecognised helper or a shadowed one, this PR deliberately does *not* rewrite, so consumer-side detection still has a job there.

<details>

**Goal given to Claude:** Fix three independent defects in this fork and land them as separate PRs (different risk profiles), each with a changeset. This is PR 1 of 3. The unquoted-`class` gap was described as the highest-value of the three, with the expected behaviour table supplied up front and already prototyped by hand.

**Explicit instructions:** don't change public API or option names; verify against the repo's own test suite and test-apps rather than by inspection; add tests for every row including the global-class and path-expression rows.

- **There is no changeset mechanism in this repo.** No `.changeset/`, no `@changesets/cli`, zero references to `changeset`. `release-plan` derives the changelog from PR labels + titles (RELEASE.md). Flagged rather than writing files the tooling would ignore; this PR wants the `bug` label.
- Verified each behaviour row against the real implementation before writing assertions, rather than predicting output.
- **Environment issue found, not caused by this PR:** `test-apps/*/package.json` have no `volta.extends`, so they default to Node 22 where `RegExp.escape` (used in `lib/path/utils.js`) does not exist and every test-app build fails. Test-apps were run via `volta run --node 24.4.0`. Reported separately.

**Round 2 — review feedback from @NullVoxPopuli (#r3722773426), "what happens if a string happens to match a class but isn't a class?"**

- Claude reproduced it rather than reasoning about it, confirmed the review was right, and found the same corruption already happened on `main` in the **quoted** form — so this PR had been propagating an existing defect to the more common spelling.
- Claude flagged that its own description had **overstated** the `classesInCss` guard as what made this safe: the guard limits blast radius but cannot tell a class name from a coincidence. Corrected.
- Claude also flagged that its own test asserting `{{if x (concat "a" "-suffix") "b"}}` renamed inside the subexpression was the weak spot, since no purely syntactic depth-based rule separates that from the `checkAlphabet` case.
- Decision, confirmed before implementing: key on *position* rather than presence in the subtree, and fix the quoted path at the same time.

**Round 3 — two follow-ups raised by the requester after discussion with @NullVoxPopuli.**

*"Make sure the allowlisted helpers are not shadowed."* Claude established by experiment, not assumption, that: a block param produces a `PathExpression` indistinguishable from the built-in; `concat` fails to compile in strict mode without an import while `if`/`unless` compile fine; `env.strictMode` is available to the AST transform but `env.locals` is empty under content-tag's eval-based scope, and `jsutils` offers no way to resolve a binding's origin. A first attempt was wrong — element block params never reached the ancestor stack — which Claude found by instrumenting the traversal after a test failed, tracing it to `All` only running for node types with no visitor of their own.

*"Does `{{concat "a" "-suffix"}}` actually work if the CSS selects `a-suffix`?"* Claude ran the template and the CSS rewriter together and confirmed it does **not**: the template emitted `a_HASH-suffix` while the CSS produced `.a-suffix_HASH`. The test in this PR had been asserting that broken behaviour. Fixed by folding literal params before renaming, with a boundary rule for dynamic params, and covered by a loose-mode browser e2e since unit tests alone never compare the emitted class against the rewritten CSS.

**Round 4 — the requester asked what `concat` does with none of this work applied.**

Claude measured `main` directly rather than describing it: there is no notion of `concat` there at all, just "rename every string literal in a quoted mustache, ignore unquoted ones". That made two things clear. `main` is *less* sound than the allowlist, since it renames literals inside any helper, shadowed or not. But it also showed that distrusting `concat` in strict mode — which Claude had implemented and flagged as the open question — would **regress a case that works on `main` today**, because the quoted form has always been handled in `.gjs`. Claude had previously described that as a lost improvement rather than a regression, which understated it. On that basis the strict-mode distrust was reverted in favour of shadow-checking alone, and a `.gjs` e2e was added for imported `concat` to pin the case that would have regressed.

The requester also asked whether `{{concat "a" (if C "-on" "-off")}}` could ideally yield `a-on_HASH`/`a-off_HASH`. Claude confirmed it is attainable — the value set is statically enumerable and concat distributes over `if` — but recommended against bundling it here, and it is recorded as a known limitation instead.

`pr-readiness` was run before each push.

</details>

### DIFF
--- a/ember-scoped-css/src/lib/rewriteHbs.classes.test.ts
+++ b/ember-scoped-css/src/lib/rewriteHbs.classes.test.ts
@@ -56,9 +56,9 @@ describe('class attribute values', () => {
 
     it('reaches literals nested in a class-building subexpression', () => {
       expect(
-        rewrite('<div class={{if x (concat "a" "-suffix") "b"}}></div>'),
+        rewrite('<div class={{if x (concat "a" " " "b") "b"}}></div>'),
       ).to.equal(
-        '<div class={{if x (concat "a_pfx" "-suffix") "b_pfx"}}></div>',
+        '<div class={{if x (concat "a_pfx" " " "b_pfx") "b_pfx"}}></div>',
       );
     });
 
@@ -131,12 +131,24 @@ describe('class attribute values', () => {
     });
 
     it('leaves a comparand alone inside a class-building helper', () => {
+      // "a" fuses with whichever branch wins, so none of these is a class name.
       expect(
         rewrite(
           '<div class={{concat "a" (if (eq this.x "b") "-on" "-off")}}></div>',
         ),
       ).to.equal(
-        '<div class={{concat "a_pfx" (if (eq this.x "b") "-on" "-off")}}></div>',
+        '<div class={{concat "a" (if (eq this.x "b") "-on" "-off")}}></div>',
+      );
+    });
+
+    it('reaches into a subexpression a boundary separates', () => {
+      // The space makes the branches whole class names rather than fragments.
+      expect(
+        rewrite(
+          '<div class={{concat "a " (if this.x "b" "global-thing")}}></div>',
+        ),
+      ).to.equal(
+        '<div class={{concat "a_pfx " (if this.x "b_pfx" "global-thing")}}></div>',
       );
     });
 
@@ -146,6 +158,129 @@ describe('class attribute values', () => {
       expect(rewrite('<div class={{someHelper "a"}}></div>')).to.equal(
         '<div class={{someHelper "a"}}></div>',
       );
+    });
+  });
+
+  /**
+   * concat joins its params into one string, so a literal is only a class name
+   * of its own when a boundary separates it from its neighbours. Postfixing a
+   * fragment would bury the postfix in the middle of the resulting class, where
+   * it matches no selector the CSS rewrite produces.
+   */
+  describe('concat fuses its params', () => {
+    /** `a-suffix` is a class in the CSS in its own right. */
+    function rewriteWithSuffixClass(hbs) {
+      return rewriteHbs(
+        hbs,
+        new Set(['a', 'b', 'a-suffix']),
+        new Set(),
+        postfix,
+      );
+    }
+
+    it('folds fused literals so the postfix lands at the end', () => {
+      expect(
+        rewriteWithSuffixClass('<div class={{concat "a" "-suffix"}}></div>'),
+      ).to.equal('<div class={{concat "a-suffix_pfx"}}></div>');
+    });
+
+    it('folds fused literals inside a branch too', () => {
+      expect(
+        rewriteWithSuffixClass(
+          '<div class={{if x (concat "a" "-suffix") "b"}}></div>',
+        ),
+      ).to.equal('<div class={{if x (concat "a-suffix_pfx") "b_pfx"}}></div>');
+    });
+
+    it('leaves fused literals alone when the fold is not a class', () => {
+      // Renaming "a" here would emit a_pfx-suffix, which matches nothing.
+      expect(rewrite('<div class={{concat "a" "-suffix"}}></div>')).to.equal(
+        '<div class={{concat "a" "-suffix"}}></div>',
+      );
+    });
+
+    it('leaves a literal that fuses with a dynamic value alone', () => {
+      expect(rewrite('<div class={{concat "a" this.x}}></div>')).to.equal(
+        '<div class={{concat "a" this.x}}></div>',
+      );
+    });
+
+    it('renames a literal a boundary separates from a dynamic value', () => {
+      expect(rewrite('<div class={{concat "a " this.x}}></div>')).to.equal(
+        '<div class={{concat "a_pfx " this.x}}></div>',
+      );
+    });
+
+    it('renames a literal that begins at a boundary', () => {
+      expect(rewrite('<div class={{concat this.x " b"}}></div>')).to.equal(
+        '<div class={{concat this.x " b_pfx"}}></div>',
+      );
+    });
+  });
+
+  describe('a shadowed helper is not our helper', () => {
+    it('skips a concat shadowed by a block param', () => {
+      expect(
+        rewrite(
+          '{{#let this.x as |concat|}}<div class={{concat "a" " " "b"}}></div>{{/let}}',
+        ),
+      ).to.equal(
+        '{{#let this.x as |concat|}}<div class={{concat "a" " " "b"}}></div>{{/let}}',
+      );
+    });
+
+    it('skips a helper shadowed by an element block param', () => {
+      expect(
+        rewrite(
+          '<Foo as |unless|><div class={{unless y "a" "b"}}></div></Foo>',
+        ),
+      ).to.equal(
+        '<Foo as |unless|><div class={{unless y "a" "b"}}></div></Foo>',
+      );
+    });
+
+    it('skips if when it is a block param', () => {
+      expect(
+        rewrite(
+          '{{#each xs as |if|}}<div class={{if y "a" "b"}}></div>{{/each}}',
+        ),
+      ).to.equal(
+        '{{#each xs as |if|}}<div class={{if y "a" "b"}}></div>{{/each}}',
+      );
+    });
+
+    it('resumes renaming once the block param is out of scope', () => {
+      expect(
+        rewrite(
+          '<Foo as |concat|></Foo><div class={{concat "a" " " "b"}}></div>',
+        ),
+      ).to.equal(
+        '<Foo as |concat|></Foo><div class={{concat "a_pfx" " " "b_pfx"}}></div>',
+      );
+    });
+  });
+
+  describe('strict mode', () => {
+    function rewriteStrict(hbs) {
+      return rewriteHbs(hbs, classes, new Set(), postfix, new Set(), true);
+    }
+
+    it('trusts if, which needs no import', () => {
+      expect(rewriteStrict('<div class={{if x "a" "b"}}></div>')).to.equal(
+        '<div class={{if x "a_pfx" "b_pfx"}}></div>',
+      );
+    });
+
+    it('trusts unless, which needs no import', () => {
+      expect(rewriteStrict('<div class={{unless x "a" "b"}}></div>')).to.equal(
+        '<div class={{unless x "a_pfx" "b_pfx"}}></div>',
+      );
+    });
+
+    it('does not trust concat, which must be imported from somewhere', () => {
+      expect(
+        rewriteStrict('<div class={{concat "a" " " "b"}}></div>'),
+      ).to.equal('<div class={{concat "a" " " "b"}}></div>');
     });
   });
 
--- a/ember-scoped-css/src/lib/rewriteHbs.classes.test.ts
+++ b/ember-scoped-css/src/lib/rewriteHbs.classes.test.ts
@@ -260,30 +260,6 @@ describe('class attribute values', () => {
     });
   });
 
-  describe('strict mode', () => {
-    function rewriteStrict(hbs) {
-      return rewriteHbs(hbs, classes, new Set(), postfix, new Set(), true);
-    }
-
-    it('trusts if, which needs no import', () => {
-      expect(rewriteStrict('<div class={{if x "a" "b"}}></div>')).to.equal(
-        '<div class={{if x "a_pfx" "b_pfx"}}></div>',
-      );
-    });
-
-    it('trusts unless, which needs no import', () => {
-      expect(rewriteStrict('<div class={{unless x "a" "b"}}></div>')).to.equal(
-        '<div class={{unless x "a_pfx" "b_pfx"}}></div>',
-      );
-    });
-
-    it('does not trust concat, which must be imported from somewhere', () => {
-      expect(
-        rewriteStrict('<div class={{concat "a" " " "b"}}></div>'),
-      ).to.equal('<div class={{concat "a" " " "b"}}></div>');
-    });
-  });
-
   describe('the scopedClass helper still collapses to a text node', () => {
     it('renames unconditionally, without a CSS lookup', () => {
       expect(
--- a/ember-scoped-css/src/lib/rewriteHbs.classes.test.ts
+++ b/ember-scoped-css/src/lib/rewriteHbs.classes.test.ts
@@ -1,0 +1,134 @@
+import { describe, expect, it } from 'vitest';
+
+import rewriteHbs from './rewriteHbs.js';
+
+const postfix = 'pfx';
+
+/**
+ * `a` and `b` are in the co-located CSS; `global-thing` deliberately is not, so
+ * every test can distinguish "renamed" from "left alone for a global class".
+ */
+const classes = new Set(['a', 'b']);
+
+function rewrite(hbs) {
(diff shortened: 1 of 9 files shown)


ember-scoped-css/src/lib/rewriteHbs.classes.test.ts
```diff
@@ -0,0 +1,134 @@
...
+      expect(
+        rewrite('<div class={{if x (concat "a" "-suffix") "b"}}></div>'),
+      ).to.equal(
+        '<div class={{if x (concat "a_pfx" "-suffix") "b_pfx"}}></div>',
+      );
+    });
+
+    it('renames every class in a multi-class literal', () => {
+      expect(rewrite('<div class={{if x "a b" "b"}}></div>')).to.equal(
+        '<div class={{if x "a_pfx b_pfx" "b_pfx"}}></div>',
```
NullVoxPopuli, line 79: what happens if a string happens to match a class but isn't a class?

```hbs
<div class={{if (checkAlphabet "a" "b") "a" "b"}}>
```
you'd want this transformed to:
```hbs
<div class={{if (checkAlphabet "a" "b") "a_pfx" "b_pfx"}}>
```

ember-scoped-css/src/lib/rewriteHbs.js
```diff
@@ -42,6 +42,21 @@ export function templatePlugin({ classes, tags, attributes, postfix }) {
     return scopedClassCandidates.some((candidate) => candidate === str);
   }
 
+  /**
+   * Only string literals are reachable: a path expression such as
+   * `{{this.fooClass}}` resolves at runtime, so it carries no class name to
+   * rename at build time.
+   */
+  function renameLiteralClasses(mustache) {
+    recast.traverse(mustache, {
```
NullVoxPopuli: traversing all StringLiterals might be dangerous

NullVoxPopuli: I see this is existing code extracted, but we do need answers to the questions above

ember-scoped-css/src/lib/rewriteHbs.classes.test.ts
adrianbw: So it turns out this is definitely a bug, and that it was also already present. I had Claude take a spin at preventing it, but it feels a bit coarse to me.
- [Here are the test cases](https://github.com/auditboard/ember-scoped-css/pull/401/changes#diff-28633c350f90a6e94854f32d9ea16a45a99cfd5bd18a77a05979db710b2eb10cR109-R150)
- [Here's the hardcoded list of helpers](https://github.com/auditboard/ember-scoped-css/pull/401/changes#diff-ebbae049b3d076c73f3d3cc54bd53478d0f2528182d64dc8919d8ccf3e5f5adcR44-R48)

Let me know what you think of that approach.

ember-scoped-css/src/lib/rewriteHbs.js
```diff
@@ -30,6 +30,58 @@ function elementHasScopedAttribute(node, attributes) {
...
+ */
+const CLASS_BUILDING_HELPERS = new Set(['concat', 'if', 'unless']);
+
+function endsAtClassBoundary(value) {
+  return /\s$/.test(value);
+}
+
+function startsAtClassBoundary(value) {
+  return /^\s/.test(value);
+}
```
adrianbw, line 68: [issue](https://pullpo.io/cc?l=issue) 
Add examples for each of these regexes.

ember-scoped-css/src/lib/rewriteHbs.js
```diff
@@ -30,6 +30,58 @@ function elementHasScopedAttribute(node, attributes) {
...
+
+/**
+ * Whether the literal at `index` is a whole class name rather than a fragment
+ * that `concat` fuses onto a neighbour. In `{{concat "a" "-suffix"}}` the "a"
+ * is a fragment: the result is one class, `a-suffix`, so postfixing "a" on its
+ * own would bury the postfix in the middle of it.
+ */
```
adrianbw, line 79: Add actual examples here.

ember-scoped-css/src/lib/rewriteHbs.js
adrianbw: Added to both predicates in 3d45da6, including the multi-class case that shows why a boundary is about the *edges* of the value rather than whether it contains whitespace at all:

```js
/**
 *   "a "     -> true
 *   "a b"    -> false, b runs on into whatever follows
 *   "a"      -> false
 *   ""       -> false
 */
function endsAtClassBoundary(value) { … }

/**
 *   " b"     -> true
 *   "a b"    -> false, a runs on from whatever precedes
 *   "b"      -> false
 *   ""       -> false
 */
function startsAtClassBoundary(value) { … }
```

The `""` row is load-bearing rather than filler: a param whose value is only known at runtime is passed through as `""`, so it contributes no boundary of its own and is bounded by its neighbours alone.

Every row was checked by calling the functions, not derived by reading the regex.

ember-scoped-css/src/lib/rewriteHbs.js
adrianbw: Added in 3d45da6, one row per shape rather than the single prose example that was there:

```js
/**
 * Whether the param at `index` is a whole class name rather than a fragment
 * that `concat` fuses onto a neighbour. Postfixing a fragment would bury the
 * postfix in the middle of the class the fragments build.
 *
 *   {{concat "a" " " "b"}}     -> all three, the spaces separate them
 *   {{concat "a" "-suffix"}}   -> neither, the result is the one class a-suffix
 *   {{concat "a" this.x}}      -> neither, "a" fuses with an unknown value
 *   {{concat "a " this.x}}     -> both, the space ends "a" and bounds this.x
 */
```

I also changed "the literal at `index`" to "the param at `index`", since the function is asked about every param including dynamic ones — which is what the last two rows are about.

The results were produced by calling `isWholeClassName` on each param of each of those four expressions, so the annotations are what the code actually returns:

```
{{concat "a" " " "b"}}     "a"=true   " "=true         "b"=true
{{concat "a" "-suffix"}}   "a"=false  "-suffix"=false
{{concat "a" this.x}}      "a"=false  this.x=false
{{concat "a " this.x}}     "a "=true  this.x=true
```

Worth noting the last row: `this.x` comes back `true`, and that is harmless rather than a bug — a bounded dynamic param is descended into, and descending into a path expression finds no literal to rename. It is only literals that a `true` actually renames.

ember-scoped-css/src/lib/rewriteHbs.classes.test.ts
```diff
@@ -0,0 +1,300 @@
...
+    function rewriteWithSuffixClass(hbs) {
+      return rewriteHbs(
+        hbs,
+        new Set(['a', 'b', 'a-suffix']),
+        new Set(),
+        postfix,
+      );
+    }
+
+    it('folds fused literals so the postfix lands at the end', () => {
```
NullVoxPopuli: if we're going to transform full strings like this, we should remove the concat (if it's the one imported from `@ember/helper` or not defined in scope (keyword))


This is happening at all is sorta scope greep tho, and I don't know that I agree with this behavior.

I kind of think we should remove all compat analyzing, and make folks use scopedClass, because there are too many ways to interpret this

ember-scoped-css/src/lib/rewriteHbs.classes.test.ts
```diff
@@ -0,0 +1,308 @@
...
+    function rewriteWithSuffixClass(hbs) {
+      return rewriteHbs(
+        hbs,
+        new Set(['a', 'b', 'a-suffix']),
+        new Set(),
+        postfix,
+      );
+    }
+
+    it('collapses a fully-fused concat to the renamed literal', () => {
```
NullVoxPopuli, line 181: if we're going to concat collapse, why not do it in any of the above tests?

As an aside, I think all the concat tests need to go in a separate file

maybe a separate file for each of if/unless/concat

ember-scoped-css/src/lib/rewriteHbs.classes.test.ts
```diff
@@ -0,0 +1,308 @@
...
+      return rewriteHbs(hbs, classes, new Set(['div']), postfix);
+    }
+
+    it('renames literals and appends the postfix class', () => {
+      expect(rewriteScopedDiv('<div class={{if x "a" "b"}}></div>')).to.equal(
+        '<div class="{{if x "a_pfx" "b_pfx"}} pfx"></div>',
+      );
+    });
+
+    it('appends the postfix class to a path expression', () => {
```
NullVoxPopuli, line 302: does this imply _all_ elements have the same class applied? regardless of if they are involved with the CSS or not?

NullVoxPopuli: is this different on main?

